PR #32, opened 2017-01-23
[COM_GROUPS][MOD_MYGROUPS][PLG_MEMBERS_GROUPS] Initial changes to add…

zooley:
… an 'archived' state to groups. @TODO: Make it mean something.

Current comparison `$group->published == 2` should be moved to an `isArchived()` method. Want to do this when we transition from old Groups class to ORM model.


core/components/com_groups/admin/controllers/manage.php
```diff
@@ -123,6 +128,11 @@ public function displayTask()
 			}
 		}
 
+		if ($this->view->filters['state'] >= 0)
+		{
+			$this->view->filters['published'] = $this->view->filters['state'];
```
zooley: This is due to the `Groups::find()` method. It currently doesn't handle array of published values. Instead, if `$filters['state]` is set *at all*, it will set the query to that value. For example, if `$this->view->filters['state'] = 0` is passed in, it'll return unpublished groups. So, I needed a way to not pass in anything so I could get all groups.

core/components/com_groups/admin/controllers/manage.php
```diff
@@ -275,9 +281,10 @@ public function saveTask()
 		}
 
 		$task = ($this->_task == 'edit') ? 'edit' : 'create';
+
 		if (!$this->authorize($task, $group))
 		{
-			return;
+			App::abort(403, Lang::txt('JERROR_ALERTNOAUTHOR'));
```
zooley: User is attempting to perform an unauthorized task (or, in other words, trying to do something they don't have permissions to do). So, a 403 seemed appropriate.

core/components/com_groups/admin/controllers/manage.php
```diff
@@ -430,10 +426,9 @@ public function saveTask()
 		}
 
 		// Output messsage and redirect
-		App::redirect(
-			Route::url('index.php?option=' . $this->_option . '&controller=' . $this->_controller, false),
-			Lang::txt('COM_GROUPS_SAVED')
-		);
+		Notify::success(Lang::txt('COM_GROUPS_SAVED'));
+
+		$this->cancelTask();
```
zooley: `cancelTask()` just redirects back to the main listing. Calling the method is shorthand for writing out the entire `App::redirect(Route::url(...));` bit. DRY

PR #64, opened 2017-01-26
[COM_GROUPS] The authorization check tries to access the not yet inst…

kevinwojo:
…antiated group var.


zooley:
Good catch.

zooley:
Didn't mean to close that. Hit the wrong button.

PR #75, opened 2017-02-17
[PURR][#1314] Using new ORCID creation workflow and upgrade API to version 2.0

kuang5:
There are two changes in this request. One is the new ORCID creation process. Another is the ORCID API upgrade.

 The existing “Create-and-claim” workflow causes millions of ORCID ID created but never claimed and used . So ORCID carries out the create-on-demand workflow to ask people register ORCID by themselves and get the ORCID Id instantly.

 The change about ORCID API is that ORCID has upgraded the API from version 1.2 to version
2.0. The API version 2.0 has been live on production since February 14 2017. 

For details please refer to ticket #1314 through link below.
https://purr.purdue.edu/support/ticket/1314?show=107&search=&limit=20&start=0

### DIFF
--- a/core/components/com_members/site/controllers/orcid.php
+++ b/core/components/com_members/site/controllers/orcid.php
@@ -147,7 +147,7 @@ class Orcid extends SiteController
 		}
 
 		$url .= implode('+AND+', $bits);
-
+		
 		$header = array('Accept: application/vnd.orcid+xml');		
 		if ($srv != 'public')
 		{
--- a/core/components/com_members/site/controllers/orcid.php
+++ b/core/components/com_members/site/controllers/orcid.php
@@ -126,7 +126,7 @@ class Orcid extends SiteController
 	{
 		$srv = $this->config->get('orcid_service', 'members');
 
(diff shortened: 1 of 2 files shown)


kevinwojo:
Hi Jerry, first pass looks okay to me. 

For future reference the Commit message tagging should looks like "[PURR][#1314] Description of the change" where [PURR] refers to the hub and [#1314] refers to the ticket number.

This helps the HUBzero team track issues.

I will have another developer review the code.

zooley:
This looks OK to me.

To add to Kevin's comments: When committing to a branch, commit messages are generally tagged with come combination of hub name, ticket number, and extension(s) affected. For example:

`[HABRI][#732][COM_TAGS] Make sure non-authorized people can't access admin tags`

If it affects multiple extensions, list them.

`[PURR][#12345][COM_TAGS][COM_SUPPORT] Example message`

If the number of extensions is so numerous that the message becomes unwieldy, asterisks can be substituted.

`[PURR][#123456][PLG_GROUPS_*] Example message`

kuang5:
Thank you Kevin and Shawn, will follow your suggestions next time.

PR #85, opened 2017-03-02
[PURR][#1301][PLG_PROJECTS_FILES] Fixes issue with removed files still being listed because the remote record did not properly remove during the sync process.

mulligaj:


core/components/com_projects/helpers/githelper.php
```diff
@@ -711,13 +711,13 @@ public function gitCheckout ($item = '' , $hash = '')
...
-	public function getFiles($subdir = '')
+	public function getFiles($subdir = '', $showAll = false)
 	{
 		// Make sure subdir has a trailing slash
 		$subdir = (!empty($subdir)) ? trim($subdir, DS) . DS : '';
-
+		$showArg = $showAll ? '-r' : '';
 		// Get Git status
-		$out = $this->callGit('ls-tree --name-only master ' . escapeshellarg($subdir));
+		$out = $this->callGit("ls-tree $showArg --name-only master " . escapeshellarg($subdir));
```
kevinwojo: I'm wondering if we replace this with a simple 'ls' ... would most of our git problems go away. That's beyond the scope of this fix, however. @zooley @zweidner

core/plugins/projects/files/helpers/sync.php
```diff
@@ -1032,6 +1032,39 @@ public function readFile($filename = '', $readAll = false)
...
+
+		foreach ($remotes['paths'] as $remote)
+		{
+			if (!in_array($remote['path'], $localIds) && $remote['converted'] != 1)
+			{
+				$objRFile->deleteRecord($projectId, $service, $remote['remote_id']);
+				$errorMessage = "Forcefully removed remote entry {$remote['path']}. \n";
+				$errorMessage .= "This is a sign that the sync did not properly work, ";
+				$errorMessage .= "therefore the cleanup step was required to remove the entry.";
+				error_log($errorMessage, 0);
```
kevinwojo: I like this. We do have a logging method which puts things in the cmsdebug.log, but for now this will do.

core/plugins/projects/files/helpers/sync.php
```diff
@@ -1032,6 +1032,39 @@ public function readFile($filename = '', $readAll = false)
 			}
 		}
 
-		return NULL;
+		return null;
+	}
+
+	/**
+	 * Check list of local files against remote files listed in the the #__project_remote_files table,
```
kevinwojo: Your docblocks are actually helpful! A breath of fresh air! :medal_sports:

kevinwojo:
Solid commit. Let's see how this does in QA!

PR #100, opened 2017-04-24
Move formatting work to API for SOLR indexing

kevinwojo:


core/components/com_groups/api/controllers/pluginsv1_0.php
```diff
@@ -83,7 +83,7 @@ class Pluginsv1_0 extends ApiController
 	 */
 	public function listTask()
 	{
-		$this->requiresAuthentication();
+		//$this->requiresAuthentication();
```
zweidner: Was this intentional?

kevinwojo: Nope ... oops

PR #115, opened 2017-06-19
[PURR][#1314]ORCID Create-on-demand workflow

kuang5:
This is ORCID create-on-demand workflow implementation. Please refer to ticket #1314 for details.


core/components/com_members/site/controllers/orcid.php
```diff
@@ -515,4 +536,119 @@ public function authorizeTask()
...
+			$update_q = "UPDATE #__user_profiles SET profile_value = " . '"' . $orcid  . '"' . " WHERE user_id = $userID AND profile_key = " . '"orcid"';
+			$db->setQuery($update_q);
+			$db->execute();
+		}
+		else
+		{
+			//Get new ID for #__user_profiles based on existing maximum ID
+			$id_q = "SELECT MAX(id) FROM #__user_profiles";
+			$db->setQuery($id_q);
+			$newID = (int)$db->loadResult() + 1;
```
zooley: You are attempting to manually increment a field that is already auto incrementing! Never, _ever_ do this. You should also be using the models for this.

```
$row = Components\Members\Models\Profile::oneByKeyAndUser('orcid', $user_id);
if (!$row->get('id')
{
    $row->set('profile_key', 'orchid');
    ... // set other fields and save
}
```

core/components/com_members/site/controllers/orcid.php
```diff
@@ -515,4 +536,119 @@ public function authorizeTask()
...
+			//Get new ordering based on existing maximum ordering
+			$order_q = "SELECT MAX(ordering) FROM #__user_profiles where user_id = $userID";
+			$db->setQuery($order_q);
+			$newOrdering = (int)$db->loadResult() + 1;
+			$orcid_sql = "INSERT INTO " . $db->quoteName('#__user_profiles') . " (" . $db->quoteName('id') . "," . $db->quoteName('user_id') . "," 
+			. $db->quoteName('profile_key') . "," . $db->quoteName('profile_value') . "," . $db->quoteName('ordering') . "," . $db->quoteName('access') 
+			. ") VALUES (" . $newID . "," . $userID . "," . '"orcid"'. "," . '"' . $orcid . '"' . "," . $newOrdering . "," . "1)";
```
zooley: None of the input has proper escaping, allowing for possible SQL injection attacks. Models use prepared statements, which guard against this.

core/components/com_members/site/controllers/orcid.php
```diff
@@ -515,4 +536,119 @@ public function authorizeTask()
...
+		if (Request::getVar('code'))
+		{
+			/*
+			* Check if there is such user's id in database. If there is, then save the orcid in #__user_profiles. 
+			* Otherwise save the orcid in session for later use.
+			*/
+			$user_id = User::get('id');
+			if ($user_id != 0)
+			{
+				Orcid::saveORCIDToProfile($user_id, $this->_userOrcidID);
```
zooley: Use `self` instead of `Orcid`.

```
self::saveORCIDToProfile(
```

core/components/com_members/site/controllers/orcid.php
```diff
@@ -515,4 +536,119 @@ public function authorizeTask()
...
+			curl_setopt($ch, CURLOPT_POSTFIELDS, $params);
+			curl_setopt($ch, CURLOPT_RETURNTRANSFER, true);
+			//Execute cURL command
+			$result = curl_exec($ch);
+			//Transform cURL response from json string to php array
+			if ($result)
+			{
+				$response = json_decode($result, true);
+				$this->_userName = $response['name'];
+				$this->_userOrcidID = $response['orcid'];
```
zooley: Minor point: Is there ever any validation that the ORCID returned is valid? That is, that it follows the pattern of an ORCID (####-####-####-####)?

core/components/com_members/site/views/orcid/tmpl/redirect.php
```diff
@@ -0,0 +1,47 @@
...
+ */
+
+// No direct access
+defined('_HZEXEC_') or die();
+
+?>
+<section class="main section">
+      <div class="grid nobreak">
+		<?php if (Request::getVar('code')) { ?>
+		<h1><?php echo Lang::txt('COM_MEMBERS_REDIRECT_ORCID_THANK_YOU'); ?><?php echo $this->userName; ?><?php echo Lang::txt('COM_MEMBERS_REDIRECT_ORCID_EXCLAMATION_MARK'); ?></h1>
```
zooley: Don't break sentences into multiple language strings. Language strings may contain variables.

```
echo Lang::txt('COM_MEMBERS_REDIRECT_ORCID_THANK_YOU', $name);

COM_MEMBERS_REDIRECT_ORCID_THANK_YOU="Thank you %s!"
```

core/components/com_members/site/controllers/orcid.php
kuang5: I tried the suggested methods above, however, that doesn't insert the profile record in #__user_profiles

kuang5: The id is still null at this moment. Is there anyway to get the incrementing id?

core/components/com_members/site/controllers/orcid.php
kuang5: No, there isn't. We just save whatever ORCID sends back. All other issues are fixed.

core/components/com_members/site/controllers/orcid.php
kevinwojo: When the object is saved, MYSQL auto-increments the ID field.

core/components/com_members/site/controllers/orcid.php
```diff
@@ -515,4 +536,119 @@ public function authorizeTask()
...
+		{
+			//Get new ID for #__user_profiles based on existing maximum ID
+			$id_q = "SELECT MAX(id) FROM #__user_profiles";
+			$db->setQuery($id_q);
+			$newID = (int)$db->loadResult() + 1;
+			//Get new ordering based on existing maximum ordering
+			$order_q = "SELECT MAX(ordering) FROM #__user_profiles where user_id = $userID";
+			$db->setQuery($order_q);
+			$newOrdering = (int)$db->loadResult() + 1;
+			$orcid_sql = "INSERT INTO " . $db->quoteName('#__user_profiles') . " (" . $db->quoteName('id') . "," . $db->quoteName('user_id') . ","
```
kevinwojo: Is there any reason in particular you have chosen to use a SQL statement rather than the Model provided for the User profile?

core/components/com_members/site/controllers/orcid.php
```diff
@@ -515,4 +537,113 @@ public function authorizeTask()
...
+	 *
+	 * @param   string   $name, $orcid
+	 * @return  void
+	 */
+	public static function saveORCIDToProfile($userID, $orcid)
+	{
+		//$userID = self::getUserID($name);
+		$db = App::get('db');
+		//Check if the user already has orcid record in #__user_profiles
+		$exist_q = "SELECT EXISTS(SELECT * FROM #__user_profiles WHERE user_id = $userID AND profile_key = " . '"orcid"' . ")";
```
kevinwojo: Use the Profile model.

core/components/com_members/site/controllers/orcid.php
```diff
@@ -515,4 +537,108 @@ public function authorizeTask()
...
+	/*
+	 * Save orcid to #_user_profiles
+	 *
+	 * @param   string   $name, $orcid
+	 * @return  void
+	 */
+	public static function saveORCIDToProfile($userID, $orcid)
+	{
+		$row = Profile::oneByKeyAndUser('orcid', $userID);
+		if (($userID == $row->get('user_id')) && ('orcid' == $row->get('profile_key')))
```
zooley: It would be more succinct and preferred to simply check if `$row` has an ID.
```
if ($row->get('id'))
```

core/components/com_members/site/controllers/orcid.php
```diff
@@ -515,4 +537,108 @@ public function authorizeTask()
...
+		{
+			$db = App::get('db');
+			$update_q = "UPDATE #__user_profiles SET profile_value = " . '"' . $orcid  . '"' . " WHERE user_id = $userID AND profile_key = " . '"orcid"';
```
zooley: This is vulnerable to SQL-injection attacks; The input does not have proper escaping. Just use the ORM model.

core/components/com_members/site/controllers/orcid.php
```diff
@@ -515,4 +537,108 @@ public function authorizeTask()
...
+			$db->setQuery($update_q);
+			$db->execute();			
+		}
+		else
+		{
+			$member = Member::oneOrNew($userID);
```
zooley: This entire `if ... else` statement is unnecessary. You already have an instantiated model with `$row`. Set the values and save. The model will handle all the necessary logic around updating an existing record or inserting a new one, escaping input, etc.

core/components/com_members/site/controllers/orcid.php
kuang5: Thank you for the advice. However, a new user doesn't have any profile record in jos_user_profiles. So $row->get('id') returns null here for a new user, which will make the statement (saving the orcid id in profile) in the if block not execute. Could you please give me any more advice?

core/components/com_members/site/controllers/orcid.php
kuang5: Good to know this, will do as the advice.

core/components/com_members/site/controllers/orcid.php
zooley: "... returns null here for a new user, which will make the statement (saving the orcid id in profile) in the if block not execute."

What? Your line of code is checking for the existence of a record. If the record is found, you performed an UPDATE. The line of code I suggested is a simpler existence check. It doesn't alter the functionality of the `if` block in any way.

kuang5: OK, here I will just check id for the existence of a profile record.

core/components/com_members/site/controllers/orcid.php
kuang5: So here the logic would be the following, it is correct?

		$row = Profile::oneByKeyAndUser('orcid', $userID);
                //If the orcid record is not found, then create such record by filling the user_id and orcid. Otherwise, just update the orcid.
		if (!($row->get('id')))
		{
			$row->set('user_id', $userID);
			
		}
		$row->set('orcid', $orcid);

kuang5: Could you please tell me when would the record be inserted in jos_user_profiles if there isn't any record for a new user and its orcid and userID are set through the profile model with $row? 
The else statement which I use can explicitly insert the orcid record in jos_user_profiles for a new user when he or she creates an ORCID after registering with PURR.

zooley: I'm not sure what you're asking here. All you need is something like the following:
```
$row = Profile::oneByKeyAndUser('orcid', $userID);

// If the record exists, you're just overwriting the existing data (user_id, profile_key, access) with the same values, so no harm done.
// If the record doesn't exist, you're setting for a new entry.
$row->set('user_id', $userID);
$row->set('access', $row->get('access', 1));
$row->set('profile_key', 'orcid');

// Set the ORCID
$row->set('profile_value', $orcid);

// Save the record
if (!$row->save())
{
    \Notify::error($row->getError());
}
```
As noted before, the model knows if it needs to perform an INSERT or an UPDATE. It also builds the SQL with prepared statements so no security issues and is more efficient.

kuang5: Thank you, Shawn. Have revised the code based on your comments.

PR #117, opened 2017-06-21
[PURR][#1423]Set resourceTypeGeneral to Collection for series

kuang5:
The DOI metadata resourceTypeGeneral is set to Collection for series. This is what has been decided by PURR team. Related ticket # is 1423.

### DIFF
--- a/core/components/com_publications/models/doi.php
+++ b/core/components/com_publications/models/doi.php
@@ -43,7 +43,7 @@ class Doi extends Object
 	/**
 	 * DOI Configs
 	 */
-	var $_configs = NULL;
+	var $_configs = null;
 
 	/**
 	 * Container for properties
@@ -57,7 +57,7 @@ class Doi extends Object
 	 *
 	 * @var object
 	 */
-	private $_db = NULL;
+	private $_db = null;
 	
 	/**
 	 * JDatabase
@@ -71,7 +71,7 @@ class Doi extends Object
 	 *
 	 * @return     void
 	 */
-	public function __construct( $pub = NULL )
+	public function __construct( $pub = null )
 	{
 		$this->_db = \App::get('db');
 
@@ -170,7 +170,7 @@ class Doi extends Object
 	 *
 	 * @return   string
 	 */
-	public function getServicePath( $doi = NULL )
+	public function getServicePath( $doi = null )
 	{
 		if (!$this->on())
 		{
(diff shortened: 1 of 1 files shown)


core/components/com_publications/models/doi.php
```diff
@@ -632,12 +639,30 @@ public function buildXml( $doi = NULL)
 			$xmlfile.='	</contributor>';
 			$xmlfile.='</contributors>';
 		}
+
+		// Get url which contains the publication id
+		$urlArr = explode('/', $this->_data["url"]);
+		// Get the content type of the publication by its id
+		$sql = "SELECT master_type FROM #__publications where id = $urlArr[4]";
```
kevinwojo: So you take a URL and don't perform ANY sanitization before injecting it into a SQL query?

core/components/com_publications/models/doi.php
```diff
@@ -57,14 +57,21 @@ class Doi extends Object
...
 	 */
-	private $_db = NULL;
+	private $_db = null;
+	
+	/**
+	 * JDatabase
+	 *
+	 * @var object
+	 */
+	const CONTENT_TYPE_SERIES = 5;
```
kevinwojo: No. We cannot make the assumption that Series is always identified by '5'

core/components/com_publications/models/doi.php
```diff
@@ -632,12 +639,30 @@ public function buildXml( $doi = NULL)
...
+
+		// Get url which contains the publication id
+		$urlArr = explode('/', $this->_data["url"]);
+		// Get the content type of the publication by its id
+		$sql = "SELECT master_type FROM #__publications where id = $urlArr[4]";
+		$this->_db->setQuery($sql);
+		$conType = $this->_db->loadResult();
+		// Set resTypeGenVal to Collection if publication is series
+		$resTypeGenVal = "Dataset";
+		if ($conType == self::CONTENT_TYPE_SERIES)
```
kevinwojo: Why is this a constant?

core/components/com_publications/models/doi.php
```diff
@@ -632,12 +639,30 @@ public function buildXml( $doi = NULL)
...
+		{
+			$resTypeGenVal = "Collection";
```
kevinwojo: This is not how the Master Types functionality works.

kevinwojo:
After further review, most of these changes are available as configuration options. The proposed changes are too specific to a single use-case. We will not accept the Pull Request on Github as we currently do not see the need to change any code. 
Each Master Type can be configured with it's own workflow. The HUBzero team has to be careful to accept changes to existing functionality that has already been established in other customers' workflows.

PR #130, opened 2017-07-25
Code style/formatting fixes

drewthoennes:
Standardizing the IDs of new items, adding colons after default fields while editing, and fixing a few incorrect titles.


core/components/com_answers/admin/views/questions/tmpl/edit.php
```diff
@@ -114,7 +114,7 @@ function submitbutton(pressbutton)
 					<tr>
 						<th><?php echo Lang::txt('COM_ANSWERS_FIELD_ID'); ?>:</th>
 						<td>
-							<?php echo $this->row->get('id', 0); ?>
```
kevinwojo: Specify the default value as an empty string? ```$this->row->get('id', '')```

drewthoennes: That makes more sense. I can make those changes now.

kevinwojo:
Hey Drew, come find me about this pull request.

PR #142, opened 2017-08-29
[PURR][#1475]Capitalize the first letter of the tag word and phrase

kuang5:
We recently reviewed the tags metadata for all published datasets and decided that all single terms and the beginning letter of the first word in a phrase should be capitalized. The change should reflect in the tags section on the publication page. The code change is for such purpose.


kuang5:
This is cancelled.

PR #143, opened 2017-09-04
Update index.php

apatsekin:
When you disable registration in Control Panel, registration icon is not hidden, but refers to 404 Page Not Found. Reason: if you change "Users -> Members -> Options -> Allow User Registration" it affects setting in com_members, not com_users.


zooley:
Good catch. I doubt any of us have bothered changing that setting in some time and simply never caught that it was looking in the wrong spot.

PR #150, opened 2017-09-14
[PURR][#1496]ORCID search process optimization

kuang5:
The ORCID search using ORCID API 2.0 and process takes long time and there are three major changes made to the existing process.

1. Exclude the single field search (first name or last searching separately) from the default search which happens when user clicks on the "Find your ID" button on ORCID section of profile page. Only search by combination of first name and last name is included in the default search. 

2. Exclude the email field from the ORCID search process since according to ORCID as of February 2017, fewer than 2% of the 3+ million email addresses on ORCID records are public, which means search by email may not return as many records as we expect.

3. Add institution name as a configurable option for search. When the option is enabled and institution name is set, the searching by combination of name and institution name becomes part of the searching process.

Please see ticket #1496 for details.


core/components/com_members/config/config.xml
```diff
@@ -283,6 +283,11 @@ HUBzero is a registered trademark of Purdue University.
 		<field name="orcid_members_token" type="text" menu="hide" default="" label="ORCID Members Token" description="Authorization token for the members ORCID service." />
 		<field name="orcid_sandbox_redirect_uri" type="text" menu="hide" default="" label="ORCID Sandbox Authentication Redirect URI" description="Authorization redirect URI for the sandbox ORCID service." />
 		<field name="orcid_members_redirect_uri" type="text" menu="hide" default="" label="ORCID Members Authentication Redirect URI" description="Authorization redirect URI for the members ORCID service." />
+		<field name="orcid_institution_field_option" type="list" default="0" label="ORCID Institution Option" description="Enable searching by institution name">
+			<option value="0">No</option>
+			<option value="1">Yes</option>
+		</field>
+		<field name="orcid_user_institution_name" type="text" menu="hide" default="Purdue University" label="Insitution name of ORCID user" description="The field for setting the insitution name." />
```
zooley: Default should be left blank. While "Purdue University" may be the default for PURR, the HUBzero platform is meant to be generic use for multiple organizations and groups.

core/components/com_members/site/controllers/orcid.php
```diff
@@ -390,77 +405,136 @@ public function fetchTask($is_return = false)
...
 		$last_name   = Request::getVar('lname', '');
-		$email       = Request::getVar('email', '');
 		$returnOrcid = Request::getInt('return', 0);
 		$isRegister  = $returnOrcid == 1;
-
+		$records = array();
+		
+		$ins_option = $this->config->get('orcid_institution_field_option');
+		
+		$ins_name = $this->config->get('orcid_user_institution_name', 'Purdue University');
```
zooley: See previous comment about default organization.

zweidner:
Committed these changes in 4d86c139377a32fdb7c632ca0cd47cd9be0a9311.

PR #166, opened 2017-10-26
[COM_WISHLIST] Fix incorrect MySQL query and incorrectly named variables

drewthoennes:


core/components/com_wishlist/site/controllers/wishlists.php
```diff
@@ -1393,6 +1393,12 @@ public function movewishTask()
 
 			if ($listid != $newlist->get('id'))
 			{
+				// Check if user is in group
+				if ($refid != 1 && !in_array(User::get('id'), \Hubzero\User\Group::getInstance($refid)->get('members')))
+				{
```
zooley: This appears to assume `$refid` is a group ID, which is not always the case. Wishlist can be associated with resources, publications, etc. This probably points more towards moving logic to individual plugins (ex: `plugins/wishlist/groups`) and calling some form of event and seeing if one of them returns false (i.e., not allowed), `Event::trigger('wishlist.onBeforeMoveWish' ...` The event trigger and parsing results could probably even be encapsulated in a method on the `Wish` model. Something like `canMoveTo($refid, $category)`.

PR #168, opened 2017-10-27
Chunked upload api

rkalyana:
Hi, 

 I've created a pull request to add a chunked upload API method that uses the same logic as the HTML5 chunked upload Javascript library (https://github.com/flowjs/flow.js) that we use in the iData Storage connector web upload page on MyGeoHub. We need this API to enable large file transfers from HydroShare into hub projects on MyGeoHub. 

Please review and let me know if any changes need to be made. 

Thanks,
Rajesh


zweidner:
Just changed the base branch so changes will be reflected moving forward.

PR #173, opened 2017-11-10
 [PURR][#1537]Fix the issue that ORCID fetch task is executed twice

kuang5:
ORCID fetchTask is executed twice after clicking on the "Search ORCID" button.

Operation procedure: Login on PURR -> Go to Profile -> ORCID -> Click on "Find your ID" -> Click on "Search ORCID". 

The reason is that there are two click event handlers are attached to the "Search ORCID" button, one is in the onClick property of the button itself and another is the registered via the HUB.Orcid.fetchOrcid method.  The latter is removed from the code to fix the issue.

### DIFF
--- a/core/components/com_members/site/assets/js/orcid.js
+++ b/core/components/com_members/site/assets/js/orcid.js
@@ -20,9 +20,16 @@ HUB.Orcid = {
 		var firstName = $('#first-name').val();
 		var lastName = $('#last-name').val();
 
-		if (!firstName && !lastName) {
-			alert('Please fill at least one of the fields.');
+		if ((!firstName) || (!lastName) || (!firstName && !lastName)) {
+			$('#alert-message').show();
 			return;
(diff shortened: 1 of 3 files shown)


kuang5:
Changes regarding "ORCID Search supports search by combination of first name and last name" are added. Please see ticket #1548.

kuang5:
The check error has nothing to do with my changes. Can you please merge what I changed or let me know what I should do with this error? Thank you.

FILE: .../hubzero-cms/core/components/com_storefront/models/Warehouse.php
----------------------------------------------------------------------
FOUND 1 ERROR AFFECTING 1 LINE
----------------------------------------------------------------------
 677 | ERROR | Control structures must have a space between the
     |       | element and its argument(s).
----------------------------------------------------------------------
Time: 1.28 secs; Memory: 22.75Mb

PR #182, opened 2017-12-05
Update users' forum categories API controller

anthonyfuentes:


core/components/com_forum/api/controllers/userscategoriesv2_0.php
```diff
@@ -189,8 +191,10 @@ protected function _requiresMatchingUser($currentUserId, $userId)
 				'status' => 'error',
 				'error' => 'User ID mismatch, unable to proceed.'
 			);
-			echo json_encode($error);
-			exit();
+
+			$error = json_encode($error);
```
zooley: `$this->send($data)` will JSON encode the var passed to it. Was it intentional to `json_encode` twice here?

PR #184, opened 2017-12-08
[PLG_MEMBERS] Fix typos, capitalize components in dropdown, and redirect to inbox a…

drewthoennes:
…fter settings save


core/plugins/members/messages/messages.php
```diff
@@ -930,7 +970,7 @@ public function savesettings($database, $option, $member)
 		// Push through to the settings view
 		$this->addPluginMessage(Lang::txt('You have successfully saved your message settings.'), 'passed');
 
-		return App::redirect(Route::url($member->link() . '&active=messages&action=settings'));
+		return App::redirect(htmlspecialchars_decode(Route::url($member->link() . '&active=messages&task=inbox&limit=' . $filters['limit'] . '&limitstart=0')));
```
zooley: Second param to `Route::url($url, $encode = true)` tells it to HTML encode special characters (ampersands) or not. This can be reduced to:
```
App::redirect(Route::url($member->link() . '&active=messages&task=inbox&limit=' . $filters['limit'] . '&limitstart=0', false));
```

PR #191, opened 2017-12-14
[CDMHUB][#1548] Move new discussion button to the left

drewthoennes:


zooley:
I'm not certain I agree with this. First, the reasoning from one person on one hub is suspect at best (there's actually research that shows left/right doesn't really have as big an impact on usability as some might think and it is largely driven by context with the rest of the UI). Also, all other controls (new category, edit category, edit post, etc) are right-aligned in the forum. Moving the "new discussion" button would actually break with that paradigm. Second, if the client insists, I think I'd rather do a CSS override for that one client rather than change it in the core for all clients.

PR #204, opened 2018-01-08
[COM_NEWSLETTER] Fixes toolbars and error in saveTask of controller

drewthoennes:


zooley:
As noted in some other pull requests, we're no longer updating 2.1. The error you fixed in saveTask() should not be present/fixed in 2.2 already. The remaining changes would need to be resubmitted on the 2.2 branch.

PR #206, opened 2018-01-12
[COM_CACHE] Fixes incorrect method call and capitalizes list items

drewthoennes:


zooley:
Some of these changes were made to the 2.2 branch. 2.1 is no longer under development. You might compare against 2.2 and see if something is missing. If so, create a new pull request on that branch.

PR #208, opened 2018-01-22
[HUBZERO][#10743] Adds license column sorting and filtering

drewthoennes:


zooley:
2.1 is no longer under development. Could you apply to 2.2 and create a new pull request? Also, be warned that com_resources underwent a rewrite in 2.2 to use ORM models. I think think it should have much affect on these changes but something to watch out for.

PR #215, opened 2018-02-07
[PURR][#1619]Support parsing the response to the DataCite DOI minting request

kuang5:
The change is to make the code be able to parse the response from the DataCite DOI minting request. The request is built with DataCite API, which is going to replace the EZID API, which is used on Hubzero platform now, by the end of year. The test result shows that the change doesn't have any side effect to the EZID API which is being used on Hubzero platform now.


zooley:
Looks like the formatting checks failed.

anthonyfuentes:
It looks like the code was updated so that the checks now pass.

PR #218, opened 2018-02-14
Mygeo 1191 master

anthonyfuentes:


core/plugins/projects/files/files.php
```diff
@@ -105,34 +107,31 @@ public function &onProjectAreas($alias = null)
...
-
+			$userIsMember = $model->access('member');
 			$active = Request::getInt('connection', 0);
 
 			$area['children'] = array();
 			$area['children'][] = array(
 				'name'  => 'default',
 				'title' => sprintf($this->params->get('default_connection_name', '%s Master Repository'), $model->get('title')),
-				'url'   => $model->link('files') . '&action=browse',
+				'url'   => UrlHelper::updatePerMembership($model->link('files') . '&action=browse', $userIsMember),
```
zweidner: Would it be better (at the cost of readability) to do this with ternary logic instead of adding a full helper?

core/plugins/projects/files/files.php
```diff
@@ -105,34 +107,31 @@ public function &onProjectAreas($alias = null)
...
 					foreach ($connections as $connection)
 					{
-						//$imgRel = '/plugins/filesystem/' . $connection->provider->alias . '/assets/img/icon.png';
-						//$img = (is_file(PATH_APP . DS . $imgRel)) ? '/app' . $imgRel : '/core' . $imgRel;
-
 						$area['children'][] = array(
 							'title' => $connection->name,
 							'name'  => $connection->provider->alias,
-							'url'   => $model->link('files') . '&action=browse&connection=' . $connection->id,
+							'url'   => UrlHelper::updatePerMembership($model->link('files') . '&action=browse&connection=' . $connection->id, $userIsMember),
```
zweidner: Same as above

core/plugins/projects/files/files.php
anthonyfuentes: That was my first thought. I opted to implement a helper to move the logic outside of files.php. I based my decision on wanting to avoid cluttering the main plugin file with logic not related to its responsibility.

It may be better to leave the logic in the main plugin file, as a ternary, to facilitate a refactoring (involving the creation of connection links/ area children) in the future.

anthonyfuentes: After changing the code to create the URLs via ternary operators located in
files.php, it does seem to me that the helper is the better way to go.
Creating the URLs via ternary operators results in cluttered code that
distracts from the primary functionality of the `&onProjectAreas` function.

PR #235, opened 2018-03-19
Habri admin edit badges master

anthonyfuentes:


core/components/com_resources/helpers/badges.php
```diff
@@ -0,0 +1,302 @@
...
+
+/**
+ * Resources badges class
+ */
+class Badges
+{
+
+	const LABEL = 'badge';
+	const TAGS_TABLE = '#__tags';
+	const BADGE_TABLE = '#__tags_object';
```
zooley: We should try to keep table names only in the models that represent them and use `Model::blank()->getTableName()` when needed elsewhere. I know there's still some legacy code in resources that directly references table names. Those need to be changed.

core/components/com_resources/helpers/badges.php
```diff
@@ -0,0 +1,302 @@
...
+	 * @param   integer  $strength    Badge strength
+	 * @return  void
+	 */
+	protected function _addBadge($tagId, $taggerId, $admin, $strength)
+	{
+		$taggedOn = Date::toSql();
+		$query = "INSERT INTO " . self::BADGE_TABLE;
+		$query .= " (objectid,tagid,strength,taggerid,taggedon,tbl,label)";
+		$query .= "values('$this->scopeId',$tagId,$strength,$taggerId,'$taggedOn','$this->scope',";
+		$query .= "'" . self::LABEL . "');";
```
zooley: Try to use models when you can instead of direct SQL. `Components\Tags\Models\Objct`, in this case.

PR #243, opened 2018-03-21
[COM_MEDIA] Updating component and removing Joomla! dependencies

drewthoennes:


drewthoennes:
Both errors are about having an opening PHP tag in .phtml files which don't need PHP tags

zooley:
We'll have to see about getting the style linter to ignore `phtml` files. In the meantime, you can get it to pass by adding the following to the top:
```
<?php
// No direct access.
defined('_HZEXEC_') or die();
?>
```

PR #250, opened 2018-03-29
Cdm 1583 pd js

anthonyfuentes:


anthonyfuentes:
I confirmed that the editor is functioning as expected by testing functionality at:

- /blog new and edit
- /groups/<group>/forum new and edit

PR #261, opened 2018-04-23
[fix] Adds column sorting, filtering, and searching

drewthoennes:


drewthoennes:
The Travis CI error is completely unrelated to this pull request. It's an issue in com_resources and everything here is in com_newsletter.

PR #268, opened 2018-05-16
[COM_TOOLS][API] add endpoint that returns tool.xml for alternate UIs

stevvvn:
We're doing some web components to support notebook (and other web) interfaces to Rappture tools. These can be generated from the already-existing tool.xml definitions, so I'd like to expose them for that purpose.

I added an optional xpath as a convenience so that components can render only relevant sections of controls.

### DIFF
--- a/core/components/com_tools/api/controllers/sessionsv1_0.php
+++ b/core/components/com_tools/api/controllers/sessionsv1_0.php
@@ -435,7 +435,7 @@ class Sessionsv1_0 extends ApiController
 			// replace all the <example> with one <examples>. each child of this is a
 			// <run source="relative_path.xml"> containing that document
 
-			// (the DOM API is compatible with SimpleXML has a better API for
+			// (the DOM API is compatible with SimpleXML & has a better API for
 			// combining documents)
 			$examples = new \DOMElement('examples');
 			$domLoader = dom_import_simplexml($loader);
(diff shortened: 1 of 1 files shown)


core/components/com_tools/api/controllers/sessionsv1_0.php
```diff
@@ -373,6 +373,49 @@ public function screenshotTask()
...
+		$tool = preg_replace('/[^-_a-z0-9]/', '', explode('/', $_SERVER['SCRIPT_URL'])[3]);
+		$path = DS . 'apps' . DS . $tool . DS . 'current' . DS . 'rappture' . DS . 'tool.xml';
+		if (!file_exists($path)) {
+			throw new Exception(Lang::txt('Unable to find tool.xml.'), 404);
+		}
+
+		$xml = file_get_contents($path);
+		$xpath = Request::getVar('xpath', '');
+
+		header('Content-type: text/xml');
```
zooley: The API defaults to returning JSON but can handle other formats. Adding format=xml to the request should format the output as XML, set the appropriate headers, etc. This seems to generate ... unexpected/mixed output format compared to the rest of the API. Would it be possible to build an object or array and send the output `$this->send($xml);` and let the API handle the output formatting?

stevvvn: I can get it into an object easily enough. `?format=xml` fails to reconstruct that object into something sufficiently resembling the original document, though.

The main issue is that SimpleXML converts `<foo bar=baz>` into `{ foo { @attributes: { bar: 'baz' } }`, so in the trip back to XML land it becomes `<foo type=object><@attributes type=object><bar type=string>baz</bar></@attributes></foo>`. It's easy enough to fix that to at least be valid, but it's no longer structurally similar to the original.

Nathan et al are coming up with a new interchange format for tool definitions that hopefully plays better in this regard by not being XML to begin with. For my current purposes, though, I just need to verbatim file or a portion of it.

stevvvn: (The task above outputs images, so there's some precedent for endpoints having baked-in types. Though I do agree it's not ideal)

stevvvn: One option might be throwing when `$_REQUEST['format'] !== 'xml'`. Forces client to confirm that other formats are not supported and they want the Rappture format specifically.

PR #270, opened 2018-05-21
 [fix] Use getter to get hzal properties

erichhuebner:


zooley:
Explicit use of the `get()` method isn't actually needed. The `Hubzero\Auth\Link` class is a relational model and the base `Hubzero\Database\Relational` class has a magic method for `__get()` that achieves the same result. So, attributes can be retrieved with either `$model->attribute` or `$model->get('attribute')`. The issue was that it had no such magic method for setting attributes and thus required explicit use of `set()`. This commit fixes that: https://github.com/hubzero/framework/commit/1c0f0f99bc76bc9648108483748f7c9e6ae3af19

PR #272, opened 2018-06-04
Purr 1644

anthonyfuentes:


core/components/com_publications/site/controllers/publications.php
```diff
@@ -657,18 +660,26 @@ public function pageTask()
 		$body = '';
 		if ($tab == 'about')
 		{
+			$publicationVersionId = $this->model->get('version_id');
+			$bundle = new Bundle([
```
zweidner: (@zooley) Should we have a configurable/killswitch on this? I know we've had issues with zip processes eating the webserver alive in the past on large publications.

PR #278, opened 2018-06-19
[feat] Implement Submit session status endpoint

anthonyfuentes:


core/components/com_tools/models/submitSession.php
```diff
@@ -0,0 +1,159 @@
...
+		if (empty($queryResults))
+		{
+			$submitSession = self::blank();
+		}
+		else
+		{
+			$submitSession = array_shift($queryResults);
+		}
+
+		return $submitSession;
```
wolzogen:
```php
return empty($queryResults) ?
    self::blank() : array_shift($queryResults);
```

core/components/com_tools/traits/modifiable.php
```diff
@@ -0,0 +1,64 @@
...
+		}
+
+		if ($this->_dataHasChanged($submittedData))
+		{
+			return Date::toSql();
+		}
+		else
+		{
+			return $this->get("modified");
+		}
```
wolzogen:
```php
return $this->_dataHasChanged($submittedData) ?
    Date::toSql() : $this->get("modified");
```

PR #280, opened 2018-06-27
Added ability to specify email addresses for new account notifications

erichhuebner:


core/plugins/user/hubzero/hubzero.php
```diff
@@ -85,10 +85,12 @@ public function onUserAfterSave($user, $isnew, $success, $msg)
 
 		// Initialise variables.
 		$config = App::get('config');
+		// Get any set emails that should be notified
+		$defs = explode(',', $this->params->get('emails', ''));
 
 		if (App::isSite())
 		{
-			if ($this->params->get('mail_to_admin', 1))
+			if ($defs[0])
```
zooley: `if (!empty($defs))` would be better here.

zooley:
Since the original configuration option was removed (`$this->params->get('mail_to_admin', 1)`), a migration is needed to ensure continued functionality. It would need to grab the params for the plugin, check if `mail_to_admin` was set and, if so, set the value of `emails` to `{config.mailfrom}`. Otherwise, after a code update, hubs that had been configured to email admins will stop emailing as `emails` would not be set.

I also wonder, if instead of a specific list of emails, one can select an access group instead? Then people that need to be notified are simply added to the necessary access group. Keeps from having email addresses littered throughout the CMS (and helps if someone changes their email).

zooley:
As per our conversation, I'll close this for now and you can re-submit an updated PR later.

PR #288, opened 2018-07-03
[feature] add sym links to publication bundle.

moligun:


core/components/com_publications/models/curation.php
```diff
@@ -2054,16 +2054,33 @@ public function showPackageContents()
 	/**
 	 * Get bundle package name
 	 *
+	 * @param	boolean	$includeVersionNum
+	 *
```
anthonyfuentes: I believe the standard docblock style is to not have a line between the params list and the return line.

core/components/com_publications/migrations/Migration20180627111315ComPublications.php
```diff
@@ -0,0 +1,82 @@
...
+			$publications = Version::all()->whereEquals('state', 1)->limit($this->limit)->start($offset);
+			foreach ($publications as $publication)
+			{
+				$pubId = $publication->get('publication_id');
+				$versionId = $publication->get('id');
+				$pubModel = new Publication($pubId, null, $versionId);
+				$pubModel->setCuration();
+				$pubModel->_curationModel->createSymLink();
+			}
+			$offset += $this->limit;
```
anthonyfuentes: Would it be safer to add `$publications->count()` to the offset instead of the limit as that number more directly reflects the number of publication records changed?

PR #293, opened 2018-07-05
[feat] Adding ability to collect extra profile data

zooley:
Each plugin now has extra parameters for each non-essential profile
field that can be retrieved upon login. Each field can be configured to
be collected or not.


anthonyfuentes:
Looks good to me.

PR #298, opened 2018-07-06
[feat] Adding a 'featured' column to projects

zooley:
This allows for marking a project as being 'featured' or not. This will
be used in front-end browsing/filtering and possibly modules.

### DIFF
--- a/core/components/com_projects/admin/controllers/projects.php
+++ b/core/components/com_projects/admin/controllers/projects.php
@@ -65,6 +65,8 @@ class Projects extends AdminController
 		$this->registerTask('applydescription', 'savedescription');
 		$this->registerTask('accesspublic', 'access');
 		$this->registerTask('accessprivate', 'access');
+		$this->registerTask('feature', 'featured');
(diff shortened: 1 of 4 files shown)


core/components/com_projects/admin/controllers/projects.php
```diff
@@ -841,6 +843,62 @@ public function unarchiveTask()
...
+		// Check for request forgeries
+		Request::checkToken(['get', 'post']);
+
+		if (!User::authorise('core.edit.state', $this->_option))
+		{
+			App::abort(403, Lang::txt('JERROR_ALERTNOAUTHOR'));
+		}
+
+		// Incoming
+		$ids = Request::getVar('id', array());
```
zweidner: Would it make more sense to use Request::getArray() to handle casting and the empty default array?

zooley: I considered that as I was updating `getVar()` to `getString()` in a number of components. It's just a change I simply hadn't gotten around to.

PR #305, opened 2018-07-16
Implement group project selective role syncing

anthonyfuentes:


core/plugins/projects/team/views/edit/tmpl/default.php
```diff
@@ -71,6 +76,15 @@
 				<span class="label-text"><?php echo Lang::txt('PLG_PROJECTS_TEAM_GROUP_LABEL_SYNC'); ?></span>
 			</label>
 
+			<select id="sync-role-selector" name="syncRole" hidden>
+				<?php foreach ($roles as $description => $value): ?>
+					<option value="" selected disabled hidden>Sync members as...</option>
+					<option value="<?php echo $value; ?>">
+						<?php echo $description; ?>
```
zooley: Ideally, this should probably be a language string. Something like `echo Lang::txt('PLG_PROJECTS_TEAM_' . $description);`

PR #317, opened 2018-07-18
 Add first_display column to host table 

lacunapremise:
This change adds a column "first_display", allowing the specification of display ranges that can be used on the execution host.  Each middleware client should be using a different range of displays (container IDs), so that multiple hubs can use the same execution host. The range is from first_display to (first_display + max_uses -1 )


core/migrations/Migration20180703151011ComTools
```diff
@@ -0,0 +1,35 @@
...
+class Migration20180703151011ComTools extends Base
+{
+	/**
+	 * Up
+	 **/
+	public function up()
+	{
+               // ADD COLUMN first_display to table host
+                if ($this->db->tableExists('host') && !$this->db->tableHasField('host', 'first_display')) {
+                        $query = "ALTER TABLE host ADD COLUMN first_display INT DEFAULT 1;";
```
zooley: This is altering a table used by the middleware/tools? The tables can be in a separate database on some hubs (e.g., nanohub) and need to use the middleware DB connection. See `core/migrations/Migration20160906100300ComTools.php` for an example.

zooley:
"may be a better fit to place it in core/components/com_tools/migrations"

Agreed. That tables are strictly tools related, so it makes sense from an organization standpoint.

lacunapremise:
I fixed the name, location, and I added the code for using the tools database.  I tested it on pdev.aws.hubzero.org.  I don't understand why the travis check failed?

lacunapremise:
Oh I see they are whitespace issues

PR #320, opened 2018-07-18
Create an index on jos_cart_meta  scope_id

lacunapremise:
Queries that specify a scope_id have caused a high database load; this index fixes the issue.


lacunapremise:
this is not the migration I was trying to get pulled.  I'll close this and try again.

PR #322, opened 2018-07-19
[fix][maint] Stricter input filtering, removing unused file

zooley:


core/components/com_answers/api/controllers/questionsv1_0.php
```diff
@@ -413,7 +413,7 @@ public function updateTask()
 
 		$tags = Request::getVar('tags', null);
 
-		if (isset($tags))
+		if (!is_null($tags) && is_string($tags))
```
zweidner: I think is_string will return false for a null variable

PR #352, opened 2018-07-26
[fix] Allow replies to project activity to show in group activity feed

moligun:
Group activity feed was not creating the proper group recipient when
it was a reply to existing activity.
The group activity feed also hid replies to comments unless
a search was performed. I removed this as it seems group activity
is often used as an indication to see if anything new has occurred,
and this includes replies to activity.
refs: https://qubeshub.org/support/ticket/1225


zooley:
Only pulling activity where `parent=0` was done on purpose. The situation is this: When a person makes a reply, it can't just show up at the top of the activity stream as it would be out of context. It needs to show up as a reply to its parent. So, when a reply is made, it trashes the recipient record of the reply's parent and creates a new one with he reply's `created` time stamp. This has the affect of moving the entire thread to the top of the activity list.

I believe the issue reported on QUBES (?) was that Project replies weren't showing up in the owning Group's feed. This is because the replies were being attached to their parent but their parent's recipient record wasn't being updated (and, thus, not being moved to the top of the activity stream). This is because the Projects feed code is only updating the Projects recipient records, instead of *all* recipient records for the parent activity. The proper fix is to change this (around line 429):

```
$past = Hubzero\Activity\Recipient::all()
	->whereEquals('scope', 'project')
	->whereEquals('scope_id', $this->model->get('id'))
	->whereEquals('log_id', $row->get('parent'))
	->whereEquals('state', 1)
	->rows();
```

... to this:

```
$past = Hubzero\Activity\Recipient::all()
	->whereEquals('log_id', $row->get('parent'))
	->whereEquals('state', 1)
	->rows();
```

And then make sure it creates new Recipient records for each Recipient it received above. (Note: I had it mark the original records as trashed and make a new record so as to preserve the original series of timestamps. Not sure if that was necessary and if the "created" date could just be changed on the original Recipient record.)

PR #356, opened 2018-07-30
[feat] Interface for managing project activity

zooley:


core/components/com_projects/admin/views/activity/tmpl/edit.php
```diff
@@ -0,0 +1,150 @@
...
+{
+	Toolbar::apply();
+	Toolbar::save();
+	Toolbar::spacer();
+}
+Toolbar::cancel();
+Toolbar::spacer();
+Toolbar::help('comment');
+?>
+<script type="text/javascript">
```
anthonyfuentes: Is it worth it to remove this inline JS to prepare for accommodating a CSP?

zooley: I'm going to in a later, multi-component update.

PR #362, opened 2018-08-02
[fix] Ensure publications don't provide duplicate search results.

moligun:
I ensured publications and resources were properly being added
to solr without producing a duplicate.

I also ensured items added to the Blacklist would be ignored when updated,
or when the index is rebuilt.

### DIFF
--- a/core/components/com_publications/api/controllers/publicationsv1_1.php
+++ b/core/components/com_publications/api/controllers/publicationsv1_1.php
@@ -157,6 +157,7 @@ class Publicationsv1_1 extends ApiController
 		}
 		elseif (isset($searchable))
 		{
+			return false;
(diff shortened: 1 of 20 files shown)


core/components/com_publications/api/controllers/publicationsv1_1.php
```diff
@@ -157,6 +157,7 @@ public function listTask()
 		}
 		elseif (isset($searchable))
 		{
+			return false;
```
zooley: I see this in a number of places. Is this not being used anymore?

moligun:
For Solr it is no longer needed, and this part was specific enough (if
admin access and has explicitly set the searchable flag, then return
results in a format solr will recognize) that having it return here
wouldn't affect any other calls to this api.

PR #375, opened 2018-08-09
[PURR][#1862]DataCite MDS API application on PURR/Hubzero

kuang5:
The DataCite MDS API is used for DOI registration. The functions in the doi.php and the invoke to them are updated in items.php and publications.php.


zooley:
Core functionality residing in a template override is not acceptable. If you need custom behavior for config fields, you can create a custom form field. Look at `core/plugins/oaipmh/resources/` for an example. Custom fields can dictate how they're rendered and push custom CSS and JS to the document, if needed.

kuang5:
Yes, the config fields need custom behavior. I looked at the code but still don't know how to implement the functionality in this way. Can you please give an example to show how the custom fields are set to push custom JS to the publication component’s configuration form?

core/components/com_publications/models/doi.php
```diff
@@ -98,9 +98,12 @@ public function configs()
 
 			$configs = new stdClass;
 			$configs->shoulder  = $params->get('doi_shoulder');
-			$configs->service   = trim($params->get('doi_service'), DS);
 			$configs->prefix    = $params->get('doi_prefix');
-			$configs->userpw    = $params->get('doi_userpw');
+			$configs->dataciteEZIDSwitch = $params->get('datacite_ezid_doi_service_switch');
```
zweidner: I think these will all return null on any system that an admin has not logged in to the administration panel and gone to the configuration for publications and saved a default value, which will cause all of your checks on the value of this to be wrong.  Perhaps @zooley  would know off the top of his head?

zooley: That is correct. Any newly introduced configuration option will return null unless the configuration has 1) been saved on the admin side, 2) a migration was created to insert a default value into the database, or 3) a default value is passed in the code (ex: `$configs->dataciteEZIDSwitch = $params->get('datacite_ezid_doi_service_switch', 'mydefault');`)

kuang5: Thank you for all of you pointing out the error.
I don't figure out the way to custom the configuration fields and push it to the core, could you please explain it in detail? The requirement is the following:
There is a list which includes three options, None, EZID, and DataCite. When the "DataCite" is selected, the DataCite DOI service is activated and only DataCite related options on the configuration form are going to show up. When "EZID" is selected, EZID DOI service is activated and only EZID related options on the configuration form are going to show up. When "None" is selected, all options related to DOI service will be hidden

zooley:
Pushing JS to the document:

```Document::addScript('/core/components/com_publications/admin/assets/js/mycustomfield.js');```

... or raw JS:

```Document::addScriptDeclaration('
    alert('foo');
');```

File: core/components/com_publications/models/doi.php (hunk 1 above)
Comment (kuang5) [reply core/components/com_publications/models/doi.php]: I will fix the error you pointed out.
This code needs to be pushed on PURR in early September, so could you please give any advise? Thank you.

Comment (zooley) [reply core/components/com_publications/models/doi.php]: I think writing a migration would be your best option. There should be a few examples of updating a component's params throughout the core.

File: core/components/com_publications/models/doi.php [hunk 2]
```diff
@@ -182,13 +186,29 @@ public function getServicePath($doi = null)
 
 		if ($doi)
 		{
-			$call  = $this->_configs->service . DS . 'id' . DS . 'doi:' . $doi;
+			if ($this->_configs->dataciteEZIDSwitch == 2)
```
zweidner: It would be nice to see appropriately named class constants used instead of integers in these flow statements.  Without digging a lot deeper into the code it's hard to tell what 0, 1, or 2 are supposed to mean.

core/components/com_publications/models/doi.php
kuang5: Could you advise if there is anything wrong regarding implementing the custom configuration parameters and custom behavior for them? Below are the solution that I come out based on online research. I don't know which core component has such feature after looking at all of the component code. 
1. Create a plugin, which includes a form that include both the existing parameters and the new parameters that I add. The plugin also has a javascript, which defines the custom behavior for publication component parameters.
2. When configuration form is prepared(onContentPrepareForm), load the custom form and push the js to the document.

kuang5: 1. Is there any standard procedure to create the migration script? 
2. Otherwise, I manually created a script file as attached. I test it by executing "muse migration --file=Migration20180821000000ComPublications.php -i -f -d=up". After the execution I go to check the database, however, the newly introduced values were not written into database. Is there any way to debug the script? I tried terminating the script but it doesn't work.

[Migration20180821000000ComPublications.txt](https://github.com/hubzero/hubzero-cms/files/2307612/Migration20180821000000ComPublications.txt)

core/components/com_publications/models/doi.php
kuang5: This is fixed. Thank you for the advise, Zach.

core/components/com_publications/models/doi.php
kuang5: I tried creating migration script by executing "./muse scaffolding create migration -e=com_publications" and it asks me to set the name and email. However, it still shows that these two options are not set after executing the settings. Please see log below.

root@example:/var/www/example# ./muse scaffolding create migration -e=com_publications

You can specify your name and email via:
  muse configuration set --user_name="John Doe"
  muse configuration set --user_email=john.doe@gmail.com

Error: failed to retrieve author name and/or email.
root@example:/var/www/example# ./muse configuration set --user_name="kuang5"
Saved new configuration!
root@example:/var/www/example# ./muse configuration set --user_name="kuang5@purd
Saved new configuration!
root@example:/var/www/example# ./muse scaffolding create migration -e=com_public

You can specify your name and email via:
  muse configuration set --user_name="John Doe"
  muse configuration set --user_email=john.doe@gmail.com

Error: failed to retrieve author name and/or email.

kuang5:
The code has been updated according to all of the comments and advises. Please review them. The DataCite service username/password are not set in the migration script and any hubs that need use DOI service have to ask PURR to get it. Thank you.

core/components/com_publications/admin/controllers/items.php
```diff
@@ -50,10 +50,19 @@
...
 	/**
-	 * Executes a task
+	 * DataCite and EZID switch options
 	 *
-	 * @return  void
+	 * @const
 	 */
+	const SWITCH_OPTION_NONE = 0;
+	const SWITCH_OPTION_EZID = 1;
+	const SWITCH_OPTION_DATACITE = 2;
```
zooley: These constants are specific to the DOI service and should be referenced from the DOI model, not added tot he controller. That would also allow them to be referenced in other areas of the code (the API, for example).

`if ($doiService->_configs->dataciteEZIDSwitch == Models\Doi::SWITCH_OPTION_DATACITE)`

core/components/com_publications/admin/controllers/items.php
```diff
@@ -533,7 +542,19 @@ public function saveauthororderTask()
 				// Update DOI
 				if (preg_match("/" . $doiService->_configs->shoulder . "/", $model->version->doi))
 				{
-					$doiService->update($model->version->doi, true);
+					if ($doiService->_configs->dataciteEZIDSwitch == self::SWITCH_OPTION_DATACITE)
```
zooley: It seems like this `if ... elseif` statement should be encapsulated in a method in the DOI model. Ideally, the rest of the code should never have references to specific DOI services; Only the DOI model cares about such things. In the controller, it should simply be `$doiService->update($model->version->doi, true);` and in that `update()` method, the DOI model determines what service to call, how to call them, etc. This also avoids the duplicated service checks throughout the controller.

kuang5:
This is fixed.

kuang5:
This is fixed as suggested. The DOI create function is also rewritten. The code changes have been committed to git.

kjgarza:
Hi guys, I think a mutual client (I am from DataCite) is trying to setup HubZero and I was wondering if this feature was in production?. I want to advise her to use this. I suppose the docs are in https://help.hubzero.org/documentation/220/managers/components/publications/datacite 

is that the case?

kuang5:
Hi Kristian,

The feature is already part of the  latest Hubzero code and the document is the reference to it.

Thank you,
Jerry

PR #383, opened 2018-08-17
[fix] Moving text into language file

zooley:


anthonyfuentes:
I noticed some inline styling in `com_support/site/views/emails/tmpl/abuse_html.php`. Is our plan to remove all inline styling in one fell swoop when the CSP is enforced?

zooley:
Those views are only used in HTML emails and the inline styling is necessary for proper display across email applications. Use of inline styles here should have no affect on/relation to the CSP policy.

PR #384, opened 2018-08-20
[fix] Change publication ftp bundle symlink to reference relative paths.

moligun:
Absolute paths to the original publication file were not accessible via FTP.
I changed the symlinks to reference the relative path instead.

refs: https://purr.purdue.edu/support/ticket/1852

### DIFF
--- a/core/components/com_publications/migrations/Migration20180820101435ComPublications.php
+++ b/core/components/com_publications/migrations/Migration20180820101435ComPublications.php
@@ -1,0 +1,112 @@
+<?php
+
+use Hubzero\Content\Migration\Base;
+
+// No direct access
+defined('_HZEXEC_') or die();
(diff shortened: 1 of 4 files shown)


core/components/com_publications/migrations/Migration20180820101435ComPublications.php
```diff
@@ -0,0 +1,112 @@
...
+	 * @var  integer
+	 */
+	public $limit = 1000;
+
+	/**
+	 * Up
+	 **/
+	public function up()
+	{
+		$versionCount = Version::all()->whereEquals('state', 1)->total();
```
zweidner: I think we generally try to avoid using models in migrations because they're not guaranteed to run in lockstep with code updates.

PR #388, opened 2018-08-22
Fix resource video

mulligaj:


zweidner:
The edit page had a call to the method whose signature had changed

PR #399, opened 2018-08-30
[feature] add courses to be searchable by solr

mulligaj:


core/components/com_courses/admin/controllers/courses.php
```diff
@@ -205,6 +205,7 @@ public function saveTask()
 
 		$tags = Request::getString('tags', '', 'post');
 		$row->tag($tags, User::get('id'));
+		$row->store(false);
```
zooley: `store()` needs to happen before `tag()` otherwise the `tag()` method won't have a course ID to link the tags to.

PR #417, opened 2018-09-10
[fix] Reshaping path variable for Dropbox

zweidner:
Needed a urldecode() in there

Fixes: https://qubeshub.org/support/ticket/1267


zweidner:
merging with #416

PR #421, opened 2018-09-12
[fix] Query for publication id when attaching the citation 

ishunko:


zooley:
I think we covered this in Slack, but this PR is trying to pull in from a fork of master when it needs to pull in from a fork of 2.2.

PR #442, opened 2018-09-24
[refactor] Adding primary keys to some tables

zweidner:
Migrations to add primary keys to tables that are missing them currently


core/components/com_tools/migrations/Migration20180703151011ComTools.php
```diff
@@ -21,7 +21,7 @@ public function up()
 		}
 
 		// ADD COLUMN first_display to table host
-		if ($mwdb->tableExists('host') && !$mwdb->tableHasField('host', 'first_display')) 
+		if ($mwdb->tableExists('host') && !$mwdb->tableHasField('host', 'first_display'))
 		{
 			$query = "ALTER TABLE host ADD COLUMN first_display INT DEFAULT 1;";
 			$this->db->setQuery($query);
```
zooley: Just noticed this migration is using a mix of `$mwdb` and `$this->db`.

zweidner: What the heck, linting commit reverted it on me....

zweidner: There we go, fixed the linting commit

PR #448, opened 2018-10-03
Added logic for rendering of Google file icons

anthonyfuentes:
Implemented approach results in code duplication between
_item_link.php & _item_link_file.php

Possible solution: implement & use Google file type subclass, which
would encapsulate Google-specific code (potentially large scope)

fixes: https://qubeshub.org/support/ticket/1297


core/plugins/projects/files/views/connected/tmpl/browse.php
```diff
@@ -74,20 +74,44 @@
 			<?php if ($this->model->access('content')) : ?>
 				<span id="manage_assets">
```
anthonyfuentes: The changes in this file are purely stylistic (line-wrapping).

PR #489, opened 2018-10-29
Update com_dataviewer download code

anthonyfuentes:
Removing output buffer functions allows CSV data to be sent to client

fixes: https://purr.purdue.edu/support/ticket/1962


core/components/com_dataviewer/site/filter/csv.php
```diff
@@ -57,7 +55,6 @@ function filter($res, &$dd, $ob_mode = false)
 		}
 	}
 	$csv .= $nl;
-	$csv .= columnsMetadata($dd);
```
zooley: Why was this line removed?

anthonyfuentes: The metadata that was added to the columns didn't seem to be anything that the user would be interested in. Without the metadata, the CSV more closely resembles the appearance of the data as seen in the dataviewer.

PR #494, opened 2018-11-01
[feat] Select specific versions of pub for a series

zooley:
Previously, selecting a publication for inclusion in a series, would
always select the most recent version. Now users can select specific
versions.

Refs: https://purr.purdue.edu/support/ticket/1965

### DIFF
--- a/core/plugins/projects/publications/views/selector/tmpl/content.php
+++ b/core/plugins/projects/publications/views/selector/tmpl/content.php
@@ -41,7 +41,8 @@ $filters = array(
 	'start'       => Request::getInt('limitstart', 0),
 	'search'      => Request::getString('search', ''),
 	'tag'         => trim(Request::getString('tag', '')),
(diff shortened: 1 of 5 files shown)


core/plugins/projects/publications/views/selector/tmpl/content.php
```diff
@@ -50,6 +51,7 @@
 $total = $model->entries('count', $filters);
 
 // Run query with limit
+$query =
```
anthonyfuentes: Will this line result in an error?

zweidner: I *think* it'll assign $query a true value (result of $results assignment).... looks like a mistake though

anthonyfuentes: I was just asking because the assignment operation was on a line by itself.

PR #505, opened 2018-11-09
Data cite mds api application

kuang5:
The fix to the issue is to register the target url and doi name when administrator approves the publication in the dataset curation process. Please check out ticket #1969. This patch should be merged in master as well. Thank you.


kuang5:
For this change, I have to check out the 2.2 branch rather than master as the base because some fixes such as admin configuration on publication component only exists in the branch 2.2 rather than on master. I have to switch back and forth from EZID to DataCite DOI service for test,

To solve the conflicts in this pull request, do you want me to keep the master changes for all files that are not modified by me?

kuang5:
I keep the changes in master and discard the changes in my files for those files that have conflicts but are not changed by me.

PR #509, opened 2018-11-12
[PURR][#1969]Fix the target url and doi name are not registered issue

kuang5:
[DataCite DOI Service] The target url and doi name is not registered when dataset is approved and published through curation workflow. The fix to the issue is to register the target url and doi name when administrator approves the publication in the dataset curation process.


core/plugins/projects/publications/publications.php
```diff
@@ -2250,15 +2250,48 @@ public function publishDraft()
...
 		{
-			$doiService->register(false, true, $pub->version->get('doi'));
-
+			// Update DOI metadata
+			$doiService->update($doi, true);
+			
 			if ($doiService->getError())
 			{
-				$this->setError($doiService->getError());
+				if ($doiService->_configs->dataciteEZIDSwitch == \Components\Publications\Models\Doi::SWITCH_OPTION_DATACITE)
```
zooley: I think I commented on this sort of action in a previous commit. Code outside of the DOI model should not know or care what service is being used. It should be entirely encapsulated in the model. If service specific messages need to be made, they should be created inside the model and passed up to outside code (example: `$msg = $doiService->getError()`).

kuang5: This issue is solved and the corresponding changes have been committed.

PR #565, opened 2018-12-11
[feat] Tell maxwell what hub is running

zweidner:


zweidner:
This should likely be left to master only for the time being

PR #575, opened 2018-12-13
[refactor] Redirect user to oauth flow immediately

zweidner:
Redirects a user to complete the setup of a connection immediately after creating it.

### DIFF
--- a/core/plugins/projects/files/connections.php
+++ b/core/plugins/projects/files/connections.php
@@ -276,6 +276,10 @@ class connections
 				'error'
 			);
 		}
+		// Instantiate the connection to trigger any oauth flow required
+		Request::setVar('connection', $connection->get('id'));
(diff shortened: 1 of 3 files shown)


core/plugins/projects/files/connections.php
```diff
@@ -276,6 +276,10 @@ public function saveconnection()
 				'error'
 			);
 		}
+		// Instantiate the connection to trigger any oauth flow required
+		Request::setVar('connection', $connection->get('id'));
+		$connection = Connection::One($connection->get('id'));
```
zooley: Method name is actually lowercase `Connection::one()`. Don't know if case mismatch will have repercussions?

zweidner: It worked when tested, something must be ucfirst-ing or tolower-ing?  Changing it though because I can't see the future :P

PR #586, opened 2019-01-02
[fix] Implement migration to add missing columns

anthonyfuentes:
It seems that a migration may have gone missing for the courses
component, so I implemented a migration to add the missing columsn to
the affected tables


anthonyfuentes:
It seems that having is_default default to NULL may cause issues. If anyone has any ideas on correctly determining a good default value for the column, please comment.

anthonyfuentes:
I created a PR that adds the helper functions implemented in the last commit to the migration base class:  https://github.com/hubzero/framework/pull/508.

PR #590, opened 2019-01-04
[feature] allow adding of custom queried search types

mulligaj:
Search Components previously only allowed indexable models to
be added and show as a main type. Allowing custom queries
will allow custom queries to be provided as a main category.


zooley:
Looks like it failed checks due to whitespace.

PR #613, opened 2019-01-18
[fix] Adding css for selects and simplifying

zooley:
Simplifying some form element CSS which keeps element sizes more
consistent. Also, minor reworking of Publication view.


zooley:
I need to tweak a few things here.

PR #640, opened 2019-02-01
[feat] Adding DOI column and editing capability

zooley:
Adding a column for DOI to the list of tool versions. This also allows
for editing some of the DOI info tied to the tool version. This does
not issue DOIs, however.

### DIFF
--- a/core/components/com_tools/admin/controllers/versions.php
+++ b/core/components/com_tools/admin/controllers/versions.php
@@ -73,61 +73,94 @@ class Versions extends AdminController
 	public function displayTask()
 	{
 		// Get Filters
-		$this->view->filters       = array();
-		$this->view->filters['id'] = Request::getState(
+		$filters       = array();
+		$filters['id'] = Request::getState(
 			$this->_option . '.' . $this->_controller . '.versions.id',
 			'id',
 			0,
 			'int'
 		);
-		$this->view->filters['sort']     = Request::getState(
-			$this->_option . '.' . $this->_controller . '.versions.' . $this->view->filters['id'] . 'sort',
+		$filters['sort']     = Request::getState(
+			$this->_option . '.' . $this->_controller . '.versions.' . $filters['id'] . 'sort',
 			'filter_order',
 			'toolname'
 		);
-		$this->view->filters['sort_Dir'] = Request::getState(
-			$this->_option . '.' . $this->_controller . '.versions.' . $this->view->filters['id'] . 'sortdir',
+		$filters['sort_Dir'] = Request::getState(
+			$this->_option . '.' . $this->_controller . '.versions.' . $filters['id'] . 'sortdir',
 			'filter_order_Dir',
 			'ASC'
 		);
-		$this->view->filters['limit']    = Request::getState(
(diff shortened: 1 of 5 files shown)


core/components/com_tools/admin/controllers/versions.php
```diff
@@ -230,6 +277,34 @@ public function saveTask()
...
+		if ($this->config->get('new_doi')
+		 && file_exists(\Component::path('com_resources') . '/models/doi.php')
+		 && substr($row->instance, -4) != '_dev')
+		{
+			require_once \Component::path('com_resources') . '/models/doi.php';
+
+			// Save DOI data
+			$dois = Request::getArray('doi', array(), 'post');
+
+			if ($dois['doi'])
```
anthonyfuentes: Will an index not found error be thrown if `$dois` is set to the provided default on line `290`? Would changing the predicate to `isset($dois['doi']) && $dois['doi']` be a good option?

zooley: An `isset()` shouldn't be necessary as the text field is on the form. If nothing was entered, the value will just be an empty string but the array key should exist regardless. It's typically only radios and checkboxes you have to do an `isset` on. And the one case where the form fields aren't available (dev version), shouldn't ever reach the save code as it checks for that.

core/components/com_tools/admin/views/versions/tmpl/display.php
```diff
@@ -43,25 +43,26 @@
...
-				</th>
-			</tr>
 			<tr>
 				<th scope="col"></th>
 				<th scope="col" class="priority-5"><?php echo Html::grid('sort', 'COM_TOOLS_COL_ID', 'id', @$this->filters['sort_Dir'], @$this->filters['sort']); ?></th>
 				<th scope="col"><?php echo Html::grid('sort', 'COM_TOOLS_COL_INSTANCE', 'toolname', @$this->filters['sort_Dir'], @$this->filters['sort']); ?></th>
 				<th scope="col" class="priority-4"><?php echo Html::grid('sort', 'COM_TOOLS_COL_VERSION', 'version', @$this->filters['sort_Dir'], @$this->filters['sort']); ?></th>
 				<th scope="col" class="priority-3"><?php echo Html::grid('sort', 'COM_TOOLS_COL_REVISION', 'revision', @$this->filters['sort_Dir'], @$this->filters['sort']); ?></th>
 				<th scope="col"><?php echo Html::grid('sort', 'COM_TOOLS_COL_STATE', 'state', @$this->filters['sort_Dir'], @$this->filters['sort']); ?></th>
+				<?php if ($this->config->get('new_doi')) { ?>
```
anthonyfuentes: Minor style comment: I recall the web team concluding we were going to use the colon and `end_` syntax for control structures in view files e.g. `if (...):`.

zooley: Changed.

core/components/com_tools/admin/controllers/versions.php
anthonyfuentes: How about the change below? Just to be safe.

`$dois = Request::getArray('doi', array('doi' => false), 'post');`

PR #666, opened 2019-02-11
[feat] Adding config option to disable comments

zooley:
Also, adding a config option to disable comment ratings.


zweidner:
:imp:

PR #679, opened 2019-02-19
[fix] Don't divide by zero

zooley:
Fixes: https://help.hubzero.org/support/ticket/11731

### DIFF
--- a/core/modules/mod_answers/tmpl/default.php
+++ b/core/modules/mod_answers/tmpl/default.php
@@ -36,7 +36,7 @@ $this->css();
 
 $total = $this->closed + $this->open;
 
-$percent = round(($this->closed / $total) * 100, 2);
+$percent = $total ? round(($this->closed / $total) * 100, 2) : 0;
 
 $this->css('
 	.' . $this->module->module . ' .graph .bar {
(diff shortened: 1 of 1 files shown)


core/modules/mod_answers/tmpl/default.php
```diff
@@ -36,7 +36,7 @@
 
 $total = $this->closed + $this->open;
 
-$percent = round(($this->closed / $total) * 100, 2);
+$percent = $total ? round(($this->closed / $total) * 100, 2) : 0;
```
zweidner: #683 I ended up using 100, not sure if it would be better to show 100% closed or 0% closed, I figured the target for closed questions would be 100?

zweidner:
Also fixes: https://hubicl.org/support/ticket/102

core/modules/mod_answers/tmpl/default.php
zooley: Hmm. I went with 0 because it would mean nothing has been closed (or opened). But, I guess you can have 100% of nothing (0% of nothing would mean you had something?).

zweidner: Probably completely arbitrary, honestly

PR #683, opened 2019-02-19
[fix] Correct divide by zero

zweidner:
Fixes: https://hubicl.org/support/ticket/102


zweidner:
Closing in favor of #679

PR #697, opened 2019-02-25
[fix] Allow different publications to cite source

anthonyfuentes:
fixes: https://purr.purdue.edu/support/ticket/1995


anthonyfuentes:
Previous query:

links_citation_query

Query for proposed changes:

links_citation_query_new

PR #699, opened 2019-02-25
[feature] Add configuration options for a Message Queue Service

mulligaj:
Added message queue options to the admin config panel to
get ready for services that will utilize the Message Queue.


zweidner:
Might be best to submit this PR against master branch

zooley:
After thinking about this for a little while and discussing with Zach, I think we're going to keep this on the master branch only. For now, at least. We can always back-port if the need arises.

PR #710, opened 2019-02-27
[fix] Putting back some needed targets

zooley:
Also, making sure needed JS libraries are loaded.

### DIFF
--- a/core/components/com_resources/admin/assets/js/media.js
+++ b/core/components/com_resources/admin/assets/js/media.js
@@ -21,28 +21,34 @@ function goUpDir()
 	frames['imgManager'].location.href = document.getElementById('imgManager').getAttribute('data-dir') + '&listdir=' + listdir.value +'&subdir='+ dir;
 }
 
-jQuery(document).ready(function ($) {
-	$('a.delete-file')
-		.on('click', function (e) {
+window.addEventListener('DOMContentLoaded', function(){
+	var files = document.getElementsByClassName('delete-file');
+
+	Array.prototype.forEach.call(files, function(el, index, array){
+		el.addEventListener('click', function(e){
 			var res = confirm($(this).attr('data-confirm'));
 			if (!res) {
 				e.preventDefault();
 			}
 			return res;
 		});
+	});
 
-	$('a.delete-folder')
-		.on('click', function (e) {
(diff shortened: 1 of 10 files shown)


core/components/com_resources/admin/assets/js/media.js
```diff
@@ -21,28 +21,34 @@ function goUpDir()
 	frames['imgManager'].location.href = document.getElementById('imgManager').getAttribute('data-dir') + '&listdir=' + listdir.value +'&subdir='+ dir;
 }
 
-jQuery(document).ready(function ($) {
-	$('a.delete-file')
-		.on('click', function (e) {
+window.addEventListener('DOMContentLoaded', function(){
+	var files = document.getElementsByClassName('delete-file');
+
+	Array.prototype.forEach.call(files, function(el, index, array){
```
anthonyfuentes: Is there a reason to use `Array.prototype.forEach.call` instead of `files.forEach` here?

core/components/com_resources/admin/assets/js/media.js
```diff
@@ -21,28 +21,34 @@ function goUpDir()
...
 			return res;
 		});
+	});
 
-	$('a.delete-folder')
-		.on('click', function (e) {
-			var res = confirm($(this).attr('data-confirm'));
+	var folders = document.getElementsByClassName('delete-folder');
+
+	Array.prototype.forEach.call(folders, function(el, index, array){
```
anthonyfuentes: Is there a reason to use `Array.prototype.forEach.call` instead of `folders.forEach`?

core/components/com_resources/admin/assets/js/media.js
zooley: `getElementsByClassName()` method doesn’t return an array (which has a `forEach()` method) — it returns an array-like object.

I think in ES6 one can also do `Array.from(items).forEach(…)`.

zooley: If ES6, I could also have done `document.querySelectorAll('.delete-file').forEach()`. _shrug_

anthonyfuentes: Interesting, and good to know.

PR #758, opened 2019-03-08
[fix] Fixing file inclusion issues

zooley:
Fixes: https://help.hubzero.org/support/ticket/11763


core/components/com_groups/admin/assets/js/groups.js
```diff
@@ -1,9 +1,16 @@
 Joomla.submitbutton = function(task) {
-	$(document).trigger('editorSave');
+	var cfrm = document.getElementById('component-form');
+console.log(cfrm);
```
anthonyfuentes: Delete `console.log`

PR #770, opened 2019-03-13
[fix] Fixing issues with checking in tables

zooley:
Also fixing an issue with marking checkedout publications as such.

### DIFF
--- a/core/components/com_checkin/models/inspector.php
+++ b/core/components/com_checkin/models/inspector.php
@@ -201,7 +201,7 @@ class Inspector extends Obj
 				continue;
 			}
 
-			$fields = $db->getTableColumns($tn, true);
+			$fields = $db->getTableColumns($tn, false);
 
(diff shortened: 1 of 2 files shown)


core/components/com_checkin/models/inspector.php
```diff
@@ -210,7 +210,7 @@ public function checkin($ids = array())
 
 			$values = array(
 				'checked_out' => 0,
-				'checked_out_time' => $fields['checked_out_time']['default']
+				'checked_out_time' => $fields['checked_out_time']->Default
```
anthonyfuentes: Is the 'D' in `Default` intentionally capitalized? Just confirming.

zooley: Yes. The property is a column from MySQL, so the query result it returns has it capitalized. Might make sense to normalize returned values to lowercase. But job for another day.

PR #832, opened 2019-03-21
[PURR][#2009]Include citation information of dataset in DOI metadata record

kuang5:
The changes include:
1. Include citation in the DOI xml when the DOI is created in the publication submission workflow.
2. Include citation for the DOIs that are set through publications on the citation admin interface when citation record is created.
3. Change the DOI registration http request method from "PUT" to "POST" according to DataCite's update on their API document.

### DIFF
--- a/core/components/com_citations/helpers/records.php
+++ b/core/components/com_citations/helpers/records.php
@@ -267,7 +267,7 @@ class Records
 			$resourceElementNodeList = $dom->getElementsByTagName("resource");
 			$sizeElementNodeList = $dom->getElementsByTagName("sizes");
 
-			if ($sizeElementNodeList->length != 0 )
+			if ($sizeElementNodeList->length != 0)
 			{
 				$relatedIdentifiersElement = $dom->createElement("relatedIdentifiers");
 				$relatedIdentifiersNode = $resourceElementNodeList->item(0)->insertBefore($relatedIdentifiersElement, $sizeElementNodeList->item(0));
@@ -363,7 +363,7 @@ class Records
 		{
 			foreach ($relatedIdentifierNodeList as $relatedIdentifier)
 			{
-				if ( ($relatedIdentifier->getAttribute("relationType") == "IsReferencedBy")
+				if (($relatedIdentifier->getAttribute("relationType") == "IsReferencedBy")
 					&& ($relatedIdentifier->getAttribute("relatedIdentifierType") == $key)
 					&& ($relatedIdentifier->nodeValue == $val))
 				{
--- a/core/components/com_citations/helpers/records.php
+++ b/core/components/com_citations/helpers/records.php
@@ -1,0 +1,410 @@
+<?php
+/**
+ * HUBzero CMS
+ *
+ * Copyright 2005-2019 HUBzero Foundation, LLC.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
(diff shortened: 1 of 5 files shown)


core/components/com_citations/helpers/records.php
```diff
@@ -0,0 +1,410 @@
...
+	 * Constructor
+	 *
+	 * @param      object $db     Database
+	 * @param      array  $config Optional configurations
+	 * @return     void
+	 */
+	public function __construct()
+	{
+		$this->_db  = \App::get('db');
+		$this->_publication_ver_tbl= '#__publication_versions';
```
zooley: Citations should not have a direct dependency on Publications. Interactions between extensions should be handled with plugins (that respond, for example, to an onAfterSave event).

core/components/com_citations/helpers/records.php
```diff
@@ -0,0 +1,410 @@
...
+	 */
+	public function getDoiList($pubIDArr, &$pubWithMultiVer)
+	{
+		$doiArr = [];
+
+		foreach ($pubIDArr as $pubID)
+		{
+			$sql = "SELECT doi FROM $this->_publication_ver_tbl WHERE publication_id = " . $pubID;
+			$this->_db->setQuery($sql);
+			$result = $this->_db->loadObjectList();
```
zooley: Avoid raw SQL where possible and use models for information retrieval.

core/components/com_citations/helpers/records.php
```diff
@@ -0,0 +1,410 @@
...
+		return $citationArr;
+	}
+
+	/**
+	 * Get DOI metadata XML from DataCite
+	 *
+	 * @param      $doi
+	 * @return     string  XML or false
+	 */
+	public function getDoiXML($doi)
```
zweidner: This method might serve a more general purpose in Components\Publications\Models\Doi... or perhaps the same task could be accomplished using Components\Publications\Models\Doi::runCurl()? (Not entirely sure on that one...)

core/components/com_citations/helpers/records.php
```diff
@@ -0,0 +1,410 @@
...
+
+	/**
+	 * Include citation and update DOI metadata record on DataCite
+	 *
+	 * @param      $xml
+	 * @param      $citationInfo
+	 *
+	 * @return     string  XML or false
+	 */
+	public function updateDoiXML($xml, $citationInfo)
```
zweidner: This doesn't look vastly different from Components\Publications\Models\Doi::dataciteMetadataUpdate(), would it be better to update that?

core/components/com_citations/helpers/records.php
kuang5: The onAfterSave is never triggered. I have installed the plugin and the status is active. Also have run the muse migration command.

core/components/com_citations/helpers/records.php
kuang5: The two functions server different purposes. The function _getDoiXML() retrieving the DOI metadata from DataCite, however, runCurl is used to register the metadata of DOI on DataCite.

core/components/com_citations/helpers/records.php
kuang5: updateDoiXML only includes the citation information in DOI metadata, however, dataciteMetadataUpdate updates all DOI's metadata.

PR #878, opened 2019-04-02
[fix] Don't have git report deleted files

zweidner:
Saves a for loop that checks if the files were deleted or not - silly
because git already would have known that information.


zooley:
Interesting bit on the `--diff-filter=ACMRTUXB`. Learn something new every day...

PR #881, opened 2019-04-03
[PURR][#2009]  Include citation information of dataset in DOI metadata record

kuang5:
The changes include:

1. Include citation for the DOIs that are set through the citation admin interface when citation record is created and is going to be associated with dataset.
2. Change the DOI registration http request method from "PUT" to "POST", which conforms to the DataCite MDS API standard.
3. Correct the option values of the context field in association table on citation component admin interface.


core/plugins/citation/doi/doi.php
```diff
@@ -0,0 +1,482 @@
...
+ * @license   http://opensource.org/licenses/MIT MIT
+ */
+
+// No direct access
+defined('_HZEXEC_') or die();
+
+/**
+ * Citation Plugin class for doi
+ */
+class plgCitationDoi extends \Hubzero\Plugin\Plugin
```
anthonyfuentes: There are a number of protected functions that should have names prefixed with an _ (lines 116, 160, 205, 284, 436).

core/plugins/citation/doi/doi.php
```diff
@@ -0,0 +1,482 @@
...
+				{
+					if ($assocParam['tbl'] == 'publication' && !empty($assocParam['doi']) && !empty($assocParam['type']))
+					{
+						$doiArr[] = $assocParam['doi'];
+						$relTypeArr[$assocParam['doi']] = $assocParam['type'];
+					}
+
+					if (!empty($assocParam['tbl']) && !empty($assocParam['oid']) && empty($assocParam['doi']) && empty($assocParam['type']))
+					{
+						Notify::warning("Both DOI and Context of publication " . $assocParam['oid'] . " are not set, so the citation won't be pushed to DataCite");
```
anthonyfuentes: Language strings should be used here to allow the plugin content to be translated to other languages. See [plg_answers_members](https://github.com/hubzero/hubzero-cms/tree/master/core/plugins/answers/members) for an example.

core/plugins/citation/doi/doi.php
```diff
@@ -0,0 +1,482 @@
...
+		curl_setopt($ch, CURLOPT_RETURNTRANSFER, true);
+		curl_setopt($ch, CURLOPT_HTTPHEADER, array('Content-Type:text/plain;charset=UTF-8'));
+		curl_setopt($ch, CURLOPT_FAILONERROR, true);
+		curl_setopt($ch, CURLOPT_CUSTOMREQUEST, 'POST');
+
+		curl_exec($ch);
+
+		$code = curl_getinfo($ch, CURLINFO_HTTP_CODE);
+
+		if ($code == 201 || $code == 200)
```
anthonyfuentes: Lines 452 to 459 can be condensed to `return $code == 201 || $code == 200;`.

core/plugins/citation/doi/doi.php
kuang5: This is fixed.

core/plugins/citation/doi/doi.php
kuang5: I used the language strings, but they don't work. Any reason for this strange thing?

core/plugins/citation/doi/doi.php
kuang5: This is fixed.

core/plugins/citation/doi/doi.php
anthonyfuentes: Language strings should work given the language files are named correctly and in the appropriate directory.

kuang5: I followed the rules to build the language. I uploaded a screenshot of the code and directory to ticket 2009: https://purr.purdue.edu/support/ticket/2009. Can you look at it and let me know if there is anything wrong?

anthonyfuentes: My apologies I forgot to mention that the plugin needs to have the property below in order for the language files to be loaded. 

`protected $_autoloadLanguage = true;`

kuang5: Thank you, the language string works now. But Notify::warning(Lang::txt('PLG_CITATION_DOI_CONTEXT_MISSING'), $assocParam['oid']); is not working, where the variable $assocParam is not taken to processe. Please advise how to solve it.

kuang5: This is solved.

core/migrations/Migration20190403124646ComCitations.php
```diff
@@ -0,0 +1,35 @@
...
+class Migration20190403124646ComCitations extends Base
+{
+	/**
+	 * Up
+	 **/
+	public function up()
+	{
+		if ($this->db->tableExists('#__citations_assoc') && !$this->db->tableHasField('#__citations_assoc', 'doiRef'))
+		{
+			$query = "ALTER TABLE `#__citations_assoc` ADD COLUMN `doiRef` VARCHAR(255) AFTER `oid`";
```
zooley: Why is this column being added?

core/plugins/citation/doi/doi.php
```diff
@@ -0,0 +1,480 @@
...
+	}
+
+	/**
+	 * Get DOI metadata XML from DataCite
+	 *
+	 * @param      array	$doi
+	 * @return     string  	XML or false
+	 */
+	protected function _getDoiXML($doi)
+	{
```
zooley: See comment from @zweidner on the first pull request: https://github.com/hubzero/hubzero-cms/pull/832/files/fc8d4104917d12cbdf1126d8e896059fb258fb13#r268175805

core/migrations/Migration20190403124646ComCitations.php
kuang5: This is used to specify the DOI whose metadata is going to include the citation information.

zooley: Isn't that the purpose of the `old` and `tbl` columns?

kuang5: old and tbl may map to multiple DOIs, so the exact DOI needs to be designated here in order to know which DOI's metadata record that the citation will be pushed to.

zooley: The correct course of action seems to be changing the oid value. That is, citations are linked to publication version IDs instead of publication IDs. Adding the DOI to the citation mapping table unnecessarily duplicates information.

kuang5: The publication version id is not available to regular citation component users who don't have access to database, so changing the oid value to publication version id will not support them using the citation application to create a citation-publication relationship. The DOI added here is only used as a target that the citation inclusion process works on, and it doesn't conflict with DOIs in other places or change their values.

zooley: Changes have been made to associate citations with version ID instead of publication ID. You will need to update your Pull Request to reflect those changes.

https://github.com/hubzero/hubzero-cms/pull/1150

kuang5: I will do it. Can you specify what is the un-straightforward to get the publication version id for a regular user such as graduate student, who does not have right to access database? You mentioned this in the joint team meeting last week.

zooley: Why would a front-end user need that? Citations are added to a publication during the contribution process. At no point in that process, to my knowledge, does the user ever enter or need to enter a specific publication or publication version ID.

kuang5: No, user doesn't. That's not the application scenario for this plugin. Through administrator interface->components->citations, user can create a citation and associate the citation to a publication. In this scenario, user does have to know the exact publication version id in order to associate the citation to the correct publication. So this is why the publication's doi is added to the web interface and the database table.

zooley: You said "... get the publication version id for a regular user such as graduate student" and then mention "Through administrator interface->components->citations". One is the front-end, one is the admin side. A regular user does not have access to the admin interface. Administrators can get the version Id from the list of versions in the admin interface--this change was in the pull request I posted.

PR #895, opened 2019-04-09
[fix][feat] Bootstrap classes, accessibility fixes

zooley:
Adding more Bootstrap class names to form elements and some
accessibility fixes while muckin' about.

### DIFF
--- a/core/components/com_wishlist/site/views/wishlists/tmpl/_comment.php
+++ b/core/components/com_wishlist/site/views/wishlists/tmpl/_comment.php
@@ -93,27 +93,39 @@ defined('_HZEXEC_') or die();
 
 						$link = $attachment->link('download');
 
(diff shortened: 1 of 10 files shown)


core/components/com_wishlist/site/views/wishlists/tmpl/_comment.php
```diff
@@ -93,27 +93,39 @@
 
 						$link = $attachment->link('download');
 
-						if ($attachment->isImage())
+						if ()
```
zweidner: if (!is_empty($link)) ?

zooley: Whoops. That was supposed to add a file exists check and the output a "file not found" block to be consistent with everywhere else (blogs, forum, etc).

PR #909, opened 2019-04-10
[maint] Moving common lang string to core file

zooley:
This moves a couple commonly repeated language strings to the main
language file to reduce duplication.


zweidner:
You'll have to rebase this one and force push it,  I don't want to flub up _comment.php

PR #935, opened 2019-04-16
Forum fix and feature

zweidner:


zweidner:
https://catalyzecare.org/support/ticket/673

PR #989, opened 2019-04-29
Hyperlink DOIs to preferred resolver

katrinleinweber:
Hello :-)

The DOI Foundation [started recommending a new resolver](https://www.doi.org/doi_handbook/3_Resolution.html#3.8). While their URL change may be a bit ironic, it's now [encrypted](https://www.ssllabs.com/ssltest/analyze.html?d=doi.org) and the old `dx` subdomain is being redirected. So, there is no urgency here.

However, for consistency, this PRs suggests to update the code that generates new DOI links.

I'm not entire sure whether search-and-replace is the correct approach for each occurrence here, though, sorry! Are any of them needed to match strings which could still contain the old version, `http://dx.doi.org`?


zweidner:
There shouldn't be any issues that I can see as far as string matching is concerned.  Thanks for the contribution :smiley:

katrinleinweber:
You're welcome, and thanks for merging :-)

PR #1006, opened 2019-05-02
[fix] Don't auto-create bundle

zooley:
Instead, display list of files for the publication.

Fixes: https://purr.purdue.edu/support/ticket/2027


core/components/com_publications/admin/views/items/tmpl/_selectcontent.php
```diff
@@ -28,6 +28,21 @@
 	// Get attachment type model
 	$attModel = new \Components\Publications\Models\Attachments($database);
 
+	foreach ($prime as $i => $elm)
```
anthonyfuentes: Since the elements' indices aren't unused, would it be better to change the loop bodies to `$elm->manifest...`?

PR #1041, opened 2019-05-08
PHP 7 compatibility fixes

ieyasu:
continue; -> continue 2; -- in PHP 7 switch is a looping structure so continue; works on the switch; the '2' goes out 2 levels to the foreach.  No idea if this still works in PHP 5.

The is_countable() prevents the error where a value that is not an array or implements Countable is passed to count().  There may be a better fix for these, but this was the most direct.

### DIFF
--- a/core/components/com_kb/site/views/articles/tmpl/article.php
+++ b/core/components/com_kb/site/views/articles/tmpl/article.php
@@ -105,7 +105,7 @@ Document::setTitle(Lang::txt('COM_KB') . ': ' . $this->category->get('title') . 
 							<a <?php if ($this->catid == $row->get('id')) { echo 'class="active" '; } ?> href="<?php echo Route::url($row->link()); ?>">
 								<?php echo $this->escape(stripslashes($row->get('title'))); ?> <span class="item-count"><?php echo $row->get('articles', 0); ?></span>
 							</a>
-							<?php if ($this->catid == $row->get('id') && count($row->children($filters)) > 0) { ?>
+							<?php if ($this->catid == $row->get('id') && is_countable($row->children($filters)) && count($row->children($filters)) > 0) { ?>
 								<ul class="categories">
 								<?php foreach ($row->children() as $cat) { ?>
 									<li>
--- a/core/modules/mod_mytickets/tmpl/default.php
+++ b/core/modules/mod_mytickets/tmpl/default.php
@@ -62,7 +62,7 @@ $this->css();
 	<h4>
 		<?php echo Lang::txt('MOD_MYTICKETS_CONTRIBUTIONS'); ?>
 	</h4>
-	<?php if (count($this->rows3) <= 0) { ?>
+	<?php if (!is_countable($this->rows3) || count($this->rows3) <= 0) { ?>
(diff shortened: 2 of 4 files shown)


core/components/com_kb/site/views/articles/tmpl/article.php
```diff
@@ -105,7 +105,7 @@
 							<a <?php if ($this->catid == $row->get('id')) { echo 'class="active" '; } ?> href="<?php echo Route::url($row->link()); ?>">
 								<?php echo $this->escape(stripslashes($row->get('title'))); ?> <span class="item-count"><?php echo $row->get('articles', 0); ?></span>
 							</a>
-							<?php if ($this->catid == $row->get('id') && count($row->children($filters)) > 0) { ?>
+							<?php if ($this->catid == $row->get('id') && is_countable($row->children($filters)) && count($row->children($filters)) > 0) { ?>
```
zooley: After taking a look at the existing code, I think a better fix might be to assign the list of children to a variable and then count and iterate over that. The underlying issue is that the `children()` method is returning a `Hubzero\Database\Relational` object. It can be iterated over (i.e., the `foreach ($row->children() ...` a couple lines down because Relational implements `IteratorAggregate`. That is, as soon as you attempt to iterator over it, PHP automatically calls its `getIterator()` method which, internally is returning a `Hubzero\Database\Rows` object.

The `count()` method doesn't work because, at that point, it's trying to count a Relational object. So, if we manually grab _that_ Rows object first, the `count` should work because `rows()` will always return a countable object (even if no records were found).

```
$children = $row->children($filters)->rows();
if ($this->catid == $row->get('id') && count($children) > 0) { ?>
	<ul class="categories">
		<?php foreach ($children as $cat) { ?>
```

core/modules/mod_mytickets/tmpl/default.php
```diff
@@ -62,7 +62,7 @@
 	<h4>
 		<?php echo Lang::txt('MOD_MYTICKETS_CONTRIBUTIONS'); ?>
 	</h4>
-	<?php if (count($this->rows3) <= 0) { ?>
+	<?php if (!is_countable($this->rows3) || count($this->rows3) <= 0) { ?>
```
zooley: I think this could be simplified to just: `<?php if (!empty($this->rows3)) { ?>` or, in the module's `helper.php`, change the two spots where it says `$this->rows3 = null;` to `$this->rows3 = array();`.

ieyasu: Yes.  That's if `<?php (empty($this->rows3)) { ?>` though (no `!`); the else is the case where there are rows of data.

core/components/com_kb/site/views/articles/tmpl/article.php
ieyasu: Aah, very good.  Calling `rows()` explicitly appears to have done the trick.

core/modules/mod_mytickets/tmpl/default.php
zooley: Ah, yes. Good catch.

zooley:
I applied the changes we talked about in a separate commit as I noticed some other minor items that needed to be adjusted at the time. I'm going to close this PR since it's out of date now but rest assured the items you pointed out have been addressed.

PR #1059, opened 2019-05-13
[fix] Can now select img for pub outside project

mdlama:
Might have been an incorrect action listed (submit)

Fixes: https://qubeshub.org/support/ticket/1420


mdlama:
I should add that I'm not entirely sure if this is the cleanest fix as I don't quite understand all of the surrounding code.  I compared the fields listed when publishing from within a project and outside a project and noticed the additional fields due to the "isProvisioned" if statement getting executed.

In any case, realized that the action listed (submit) might not be the correct one so I removed it.  It appears to work.

PR #1067, opened 2019-05-15
[fix] Ensure link is removed if master type blacklisted.

mulligaj:
The cron job wasn't removing ftp file links if the file
was of a master type that didn't require a bundle.

fixes: https://purr.purdue.edu/support/ticket/2039


core/plugins/cron/publications/publications.php
```diff
@@ -479,22 +479,23 @@ public function updateFtpLinks(\Components\Cron\Models\Job $job)
...
 			->whereEquals('#__publication_versions.state', 1)
-			->whereNotIn('t.alias', array('series', 'databases'))
 			->where('#__publication_versions.published_up', '>=', $yesterday, 'AND', 1)
 			->where('#__publication_versions.published_up', '<=', $today, 'AND', 1)
 			->resetDepth()
 			->where('#__publication_versions.published_down', '>=', $yesterday, 'OR', 1)
 			->where('#__publication_versions.published_down', '<=', $today, 'AND', 1)
 			->rows();
 
+		$typeBlackList = array('series', 'databases');
```
zooley: Thought: Make this is customizable param on the task where the user can select the publication types to ignore? Maybe a job for another day.

zweidner: I think I would second both of those... is this something that could be encoded into the master type itself as a param or column via migration?

mulligaj: I added this as a configuration in the same spot as the other FTP configuration within the Publication params, as well as added a Migration to automatically set the default of this field to be "series, databases".

PR #1084, opened 2019-05-22
[fix] Fixes jquery .load() issues after update

mdlama:
Fixes: https://qubeshub.org/support/ticket/1424

### DIFF
--- a/core/components/com_groups/site/assets/js/groups.js
+++ b/core/components/com_groups/site/assets/js/groups.js
@@ -242,7 +242,7 @@ HUB.Groups = {
 
 			//load a refreshed version of the logo list
 			//reset option previously selected by user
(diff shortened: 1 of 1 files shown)


zooley:
See https://github.com/hubzero/hubzero-cms/commit/45574c39d58f99ad0aad3d2adc9e673fe49c236d

I had initially already made the changes in your commit and they ere reverted with he above commit. See the commit message for the reason. As such, I'm not sure how your change fixed things as it should have actually broken them (if I'm reading everything correctly). I just tested my dev machine with the `.load()` method and everything worked.

mdlama:
Huh, okay.  Well, as long as you already have a fix!  Thanks Shawn.  I'll close this down...

PR #1114, opened 2019-06-03
replace create_function() with new anonymous function syntax

ieyasu:
As of PHP 7.2, [create_function() is deprecated](https://www.php.net/manual/en/function.create-function.php).  I've replaced all instances I could find and test with anonymous functions.  This should be compatible with >= 5.3.0 (see link).  I didn't see any guidance on formatting anonymous functions so I went with the style I've seen in Javascript.

I noticed that geshi also has an instance of create_function().  There is a newer version out, 1.0.9.0, but it hasn't gotten rid of create_function().  So you know.  It's an external dependency so I figured I shouldn't touch it.

### DIFF
--- a/core/components/com_search/models/basic/terms.php
+++ b/core/components/com_search/models/basic/terms.php
@@ -326,7 +326,10 @@ class Terms extends Obj
 		usort($chunks, function($a, $b) {
 			$al = strlen($a);
 			$bl = strlen($b);
-			if ($al == $bl) return 0;
+			if ($al == $bl)
+			{
+				return 0;
+			}
 			return $al > $bl ? -1 : 1;
 		});
 		return '(' . join('|', array_map('preg_quote', $chunks)) . '[[:alpha:]]*)';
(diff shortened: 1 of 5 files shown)


core/components/com_search/models/basic/terms.php
```diff
@@ -323,7 +323,12 @@ private function add_chunk(&$partial, &$sign = '', $quoted = false)
 	public function get_word_regex()
 	{
 		$chunks = $this->get_stemmed_chunks();
-		usort($chunks, create_function('$a, $b', '$al = strlen($a); $bl = strlen($b); if ($al == $bl) return 0; return $al > $bl ? -1 : 1;'));
+		usort($chunks, function($a, $b) {
+			$al = strlen($a);
+			$bl = strlen($b);
+			if ($al == $bl) return 0;
```
zooley: Style linter is complaining about this. Inline control structures aren't allowed. It needs to be:

```
if ($al == $bl)
{
    return 0;
}
```

zooley:
Aside from the one line being flagged by the style linter, it looks good. To be honest, I wasn't even aware `create_function()` was being used anywhere.

As for Geshi, it's definitely an older library. There may be better options out there. I'll have to do some research.

core/components/com_search/models/basic/terms.php
ieyasu: Okay.  I finally got phpcs up and running so I should be able to do these checks in the future.

PR #1115, opened 2019-06-03
PHP 7 compatibility

ieyasu:
This is another 

> count(): Parameter must be an array or an object that implements Countable'

 error fix.  There's probably a cleaner way to write it, but this is the safest thing that worked for me.

### DIFF
--- a/core/plugins/search/members/members.php
+++ b/core/plugins/search/members/members.php
@@ -229,7 +229,7 @@ class plgSearchMembers extends \Hubzero\Plugin\Plugin
 				u.access IN (" . implode(',', User::getAuthorisedViewLevels()) . ") AND " . join(' AND ', $addtl_where)
 		);
 		$assoc = $sql->to_associative();
-		if (!count($assoc))
+		if (!is_array($assoc) || !count($assoc))
 		{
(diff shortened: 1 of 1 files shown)


zooley:
It seems to be that `Components\Search\Models\Basic\Result\Sql()` class, line 56, is returning a `Components\Search\Models\Basic\Result\Blank` object instead of `Components\Search\Models\Basic\Result\AssocList` (which implements `Iterator`). It think the proper fix might be to change line 56 of `Components\Search\Models\Basic\Result\Sql()` to `$rows = array();`. This will then fall through to line 58, which returns the proper `AssocList` object. I'll have to test some.

ieyasu:
Using `$rows = array();` in Components\Search\Models\Basic\Result\Sql() line 56  wasn't enough.  It still complained about $assoc not being Countable.   If I implement Countable on AssocList with the count() function

    public function count()
    {
        return count($this->rows);
    }

it does work.  It's all about that Countable interface.  I can issue a second pull request if that would help.

zooley:
Good catch. I didn't look close enough at the `AssocList` model.

PR #1153, opened 2019-06-24
[PURR][#2009] Include citation information of dataset in DOI metadata record

kuang5:
The changes include:

1. Create a doi plugin of citation component for pushing the citation to the corresponding DOI's metadata record on DataCite.
2. Correct the option values of the context field on citation component admin interface. 
3. Change the DOI registration http request method from "PUT" to "POST", which conforms to the DataCite MDS API standard.

### DIFF
--- a/core/components/com_citations/admin/views/citations/tmpl/edit.php
+++ b/core/components/com_citations/admin/views/citations/tmpl/edit.php
@@ -330,8 +330,8 @@ $journal = (!preg_match('!\S!u', $journal)) ? utf8_encode($journal) : $journal;
 								<td>
 									<select name="assocs[<?php echo $i; ?>][type]" class="noUniform">
 										<option value=""<?php echo ($assocs[$i]->type == '') ? ' selected="selected"': ''; ?>><?php echo Lang::txt('SELECT'); ?></option>
-										<option value="References"<?php echo ($assocs[$i]->type == 'References') ? ' selected="selected"': ''; ?> ><?php echo Lang::txt('COM_CITATIONS_CONTEXT_REFERENCES'); ?></option>
-										<option value="IsReferencedBy"<?php echo ($assocs[$i]->type == 'IsReferencedBy') ? ' selected="selected"': ''; ?>><?php echo Lang::txt('COM_CITATIONS_CONTEXT_REFERENCEDBY'); ?></option>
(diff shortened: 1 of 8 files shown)


core/components/com_citations/admin/views/citations/tmpl/edit.php
```diff
@@ -330,8 +330,8 @@
 								<td>
 									<select name="assocs[<?php echo $i; ?>][type]" class="noUniform">
 										<option value=""<?php echo ($assocs[$i]->type == '') ? ' selected="selected"': ''; ?>><?php echo Lang::txt('SELECT'); ?></option>
-										<option value="references"<?php echo ($assocs[$i]->type == 'references') ? ' selected="selected"': ''; ?>><?php echo Lang::txt('COM_CITATIONS_CONTEXT_REFERENCES'); ?></option>
-										<option value="referencedby"<?php echo ($assocs[$i]->type == 'referencedby') ? ' selected="selected"': ''; ?>><?php echo Lang::txt('COM_CITATIONS_CONTEXT_REFERENCEDBY'); ?></option>
+										<option value="References"<?php echo ($assocs[$i]->type == 'References') ? ' selected="selected"': ''; ?> ><?php echo Lang::txt('COM_CITATIONS_CONTEXT_REFERENCES'); ?></option>
```
zooley: If you change these values, you'd also need to write a migration to update everything in the database and change any code that may reference those values (e.g., pig_resources_citations, pig_publications_citations, plg_oaipmh_publications, plg_oaipmh_resources, etc). I would recommend *not* changing the values. I understand that you probably changed them to match DC terms but the existing codebase is already setup to handle the old values and translate to DC where needed.

kuang5: No problem. I committed the changes.

core/plugins/citation/doi/doi.php
```diff
@@ -0,0 +1,487 @@
...
+	 * @var object
+	 */
+	public $_configs = null;
+
+	/**
+	 * DataCite and EZID switch options
+	 *
+	 * @const
+	 */
+	const SWITCH_OPTION_NONE = 0;
```
zooley: These constants are duplicated from `Components\Publications\Models\Doi`. It'd be preferred to include the model and use `Components\Publications\Models::SWITCH_OPTION_DATACITE` instead of `self::SWITCH_OPTION_DATACITE`.

kuang5: This has been updated.

PR #1158, opened 2019-06-26
[fix] Updating for changes in Twitter's embed code

zooley:
Refs: https://twittercommunity.com/t/deprecating-widget-settings/102295


core/modules/mod_twitterfeed/tmpl/default.php
```diff
@@ -24,7 +24,7 @@
 $widgetSettings .= ($this->params->get('displayFooter') == 'no') ? ' nofooter' : '';
 $widgetSettings .= ($this->params->get('displayBorders') == 'no') ? ' noborders' : '';
 
-$this->js();
```
zweidner: Is there any reason to avoid deleting ../assets/js/mod_twitterfeed.js at this time?

zooley: No. I paused on that out of wondering if the external Twitter script could replace it but didn't end up testing that and simply forgot to delete the local JS.

PR #1168, opened 2019-07-02
The others variable needs to be an instance variable

ieyasu:
Presumed to be a PHP 7 compatibility bug, though I'm not sure how it worked properly before.  $others is initialized to an array right above a loop working with $this->others.  The default_navigation.php file also refers to $this->others.  The fix is to change the local variable to $this->others.


zooley:
This was already fixed in the 2.2 branch with this commit:

https://github.com/hubzero/hubzero-cms/commit/2732bdd61a98f6c93871b7bf35534a8ca8f64127#diff-c8f7e6b9092786924abfde7309cbac60

I simply didn't get around to cherry-picking it over to master until now.

PR #1183, opened 2019-07-17
[fix] Use publication version ID

zooley:


anthonyfuentes:
Ticket: https://purr.purdue.edu/support/ticket/2066

PR #1201, opened 2019-08-01
Update group's download code to be more specific

anthonyfuentes:


anthonyfuentes:
Fixes: https://stemedhub.org/support/ticket/629

PR #1212, opened 2019-08-08
[fix] Properly join tables to allow showing tags properly

ishunko:
Rewrite the table joins to properly displayed tagged publications.

### DIFF
--- a/core/components/com_publications/tables/publication.php
+++ b/core/components/com_publications/tables/publication.php
@@ -1,40 +1,15 @@
 <?php
 /**
- * HUBzero CMS
- *
- * Copyright 2005-2015 HUBzero Foundation, LLC.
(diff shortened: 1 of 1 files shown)


core/components/com_publications/tables/publication.php
```diff
@@ -1,8 +1,33 @@
 <?php
 /**
- * @package    hubzero-cms
- * @copyright  Copyright 2005-2019 HUBzero Foundation, LLC.
- * @license    http://opensource.org/licenses/MIT MIT
+ * HUBzero CMS
+ *
```
anthonyfuentes: It seems like you may have been working on an older version of this file as the expanded copyright notice is present.

Are all of the code changes below correct?

PR #1224, opened 2019-08-21
Update flot saveimage master

ishunko:


anthonyfuentes:
The commits in this PR don't match https://github.com/hubzero/hubzero-cms/pull/1223. I'm guessing this should be redone?

PR #1230, opened 2019-09-03
Plg domain restriction master

ishunko:


anthonyfuentes:
The newline characters of `core/plugins/user/domainrestriction/assets/js/base64.js` need to be converted from "\r\n" to "\n".

I would recommend running `$ dos2unix base64.js`. Let me know if you need any help.

PR #1231, opened 2019-09-04
Prevent merging tags code creating multiple associatins between the s…

ishunko:
…ame object and the same tag

Also a migration to create a unique index


anthonyfuentes:
Given the license header reverted, it seems like you were working off an old branch for this PR. Can you confirm that all of the changes outside of the header are as intended?

PR #1247, opened 2019-09-17
Update external link rel attributes

anthonyfuentes:
fixes: https://help.hubzero.org/support/ticket/11902


anthonyfuentes:
The files that are causing the checks to fail are ok to ignore.

PR #1255, opened 2019-09-20
Add user agent to avoid returning bad URL from FaceBook

ishunko:


anthonyfuentes:
Would it be best to try to source the user's user agent before defaulting to `'Mozilla/5.0 (Macintosh; Intel Mac OS X 10_13_6) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/77.0.3865.90 Safari/537.36'`?

One option for sourcing a user agent is `$_SERVER['HTTP_USER_AGENT'] `.

PR #1264, opened 2019-10-07
Fix the issue that wrong tags or null returned in OAI-PMH query result

kuang5:
The OAI-PMH metadata record harvesting returns wrong tags for the dataset/publication or does not return any tags in metadata record. Details in ticket: https://purr.purdue.edu/support/ticket/2078


anthonyfuentes:
Hello Jerry. Can you please mirror this pull request to the CMS master branch? Let me know if you have any questions. Thanks.

PR #1266, opened 2019-10-07
Solr boost update master

anthonyfuentes:


anthonyfuentes:
I believe that we can safely ignore the failed check.

PR #1294, opened 2019-11-07
Add the id to the citation section to make the anchor at the top of t…

ishunko:
…he page work


anthonyfuentes:
Hello Ilya. It just occurred to me that the license block in the changed file seems to be the old version. Can you confirm that all of the code changes in the file are intended and nothing other than the license block was reverted?

PR #1300, opened 2019-11-08
Catch abuse email errors master

ishunko:


anthonyfuentes:
Hello Ilya. This PR seems to have extraneous commits, so I created a PR with only the relevant changes here: https://github.com/hubzero/hubzero-cms/pull/1300.

PR #1310, opened 2019-11-18
File icon size fix master

ishunko:


anthonyfuentes:
See https://github.com/hubzero/hubzero-cms/pull/1311

PR #1361, opened 2020-02-11
Localize group calendar event times

anthonyfuentes:
fixes: https://qubeshub.org/support/tickets/1673

### DIFF
--- a/core/plugins/groups/calendar/helpers/userLocalizer.php
+++ b/core/plugins/groups/calendar/helpers/userLocalizer.php
@@ -35,13 +35,19 @@ class UserLocalizer
 
 	protected function _getUserTimezone()
 	{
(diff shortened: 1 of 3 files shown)


core/plugins/groups/calendar/helpers/userLocalizer.php
```diff
@@ -0,0 +1,53 @@
...
+	{
+		$userParams = $this->_getUserParams();
+
+		return Arr::getValue($userParams, 'timezone', $this->systemTimezone);
+	}
+
+	protected function _getUserParams()
+	{
+		$userId = User::get('id');
+		$query = "SELECT params from jos_users where id = $userId;";
```
hubzero-admin: Escape $userId

PR #1363, opened 2020-02-12
Integretion of new publications

datafatmunger:
* Adds a new button to create a new style publication in the projects plugin.
* Adds a new link next to the list of publications to edit as a new publication in the projects plugin.
* Adds admin option to toggle new publications component.


ishunko:
Doing it!

PR #1364, opened 2020-02-14
Add exceptions to let the unapproved users submit tickets: CDM 1910

ishunko:


anthonyfuentes:
I discussed this change with Nate, and we agreed that this should be a configurable option.

hubzero-admin:
Given guest users can create tickets at present, a wider set of permission configurations will be added.

PR #1366, opened 2020-02-18
don't copy tags over if item already has a tag

datafatmunger:
https://nanohub.org/support/ticket/363916

### DIFF
--- a/core/components/com_tags/models/objct.php
+++ b/core/components/com_tags/models/objct.php
@@ -231,6 +231,17 @@ class Objct extends Relational
 	}
 
 	/**
+	 * Checks if a particular objct has a tag
+	 *
+	 * @param   integer  $tagid  ID of tag to check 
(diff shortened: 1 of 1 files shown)


core/components/com_tags/models/objct.php
```diff
@@ -230,6 +230,17 @@ public static function moveTo($oldtagid=null, $newtagid=null)
...
 
+	/**
+	 * Checks if a particular objct has a tag
+	 *
+	 * @param   integer  $tagid  ID of tag to check 
+	 * @return  boolean  True if Objct has tag 
+	 */
+	public static function hasTag($tagid=null)
+	{
+		return (bool)($tagid && self::all()->whereEquals('tagid', $tagid)->rows());
```
hubzero-admin: Check if `->rows` is always truthy b/c it returns an object?

PR #1367, opened 2020-02-18
don't copy tags over if item already has a tag

datafatmunger:
https://nanohub.org/support/ticket/363916


anthonyfuentes:
Child PR:
https://github.com/datafatmunger/hubzero-cms/pull/1

PR #1372, opened 2020-02-26
Convert tags to an array to find the new tags added

ishunko:


hubzero-admin:
Compare with https://github.com/hubzero/hubzero-cms/pull/1380/files

PR #1377, opened 2020-02-29
[fix] Fixes incorrect argument on stripslashes

mdlama:
Refs: 32ac61f


mdlama:
Should fix https://qubeshub.org/support/ticket/1718

mdlama:
(although I've already got a patch in our app version but am currently unable to pull into production due to https://qubeshub.org/support/ticket/1690)

PR #1380, opened 2020-03-05
Update tag name lookup query

anthonyfuentes:
The previous query returned an array of arrays, so the in array check
would fail for any associated tags

fixes: https://nanohub.org/support/ticket/363378


hubzero-admin:
https://github.com/hubzero/hubzero-cms/pull/1372

PR #1382, opened 2020-03-06
Check if the evaluated code is indeed JS

ishunko:


hubzero-admin:
Look into possible security concerns.

PR #1405, opened 2020-04-07
Add SVG icons definitions support for templates. 

ishunko:
Add support comments copy link functionality.

This adds SVG sprite definitions functionality to the default template and it is being used by support comments.

All templates need to be modified to take advantage of the SVG sprites. Each template needs to add

/templates/xxx/img/icon-svg-defs.php (core/templates/kimera/img/icon-svg-defs.php) and include it in the beginning of the body tag.
Also, need to add CSS after including the core utilities.less (to make the icon SVG visible):

.cond-svg-icons-defs {
	display: flex;
}


core/components/com_support/site/assets/js/ticket.js
```diff
@@ -0,0 +1,41 @@
+jQuery(document).ready(function($) {
+	var copyBobBtn = document.querySelector('.js-copy-bob-btn'),
+		copyJaneBtn = document.querySelector('.js-copy-jane-btn');
```
hubzero-admin, line 3: Remove sample code.

PR #1418, opened 2020-05-12
Fixes Comhub collection hash file download failure

erichhuebner:
sha256 hash creates a URL greater than 255 characters on the communityhub.purdue.edu and attempting to download a file results failure (0Kb file).

### DIFF
--- a/core/components/com_collections/models/asset.php
+++ b/core/components/com_collections/models/asset.php
@@ -237,10 +237,12 @@ class Asset extends Base
 	 */
 	public function link($size = 'original')
 	{
-		$path  = $this->filespace() . DS . $this->get('item_id') . DS;
+		$base = Request::base();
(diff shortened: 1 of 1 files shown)


core/components/com_collections/models/asset.php
```diff
@@ -237,10 +237,12 @@ public function height()
 	 */
 	public function link($size = 'original')
 	{
-		$path  = $this->filespace() . DS . $this->get('item_id') . DS;
+		$base = Request::base();
+		Log::debug($base);
```
hubzero-admin, line 241: @erichhuebner to:

- [ ] remove debug log
- [ ]  consider `PATH_APP`

erichhuebner:
PATH_APP can't be used as it provides the path to the component and not the URL string that is needed.

PR #1422, opened 2020-05-13
Add local images and math support to MD parser

ishunko:


core/plugins/wiki/parsermarkdown/parser.php
```diff
@@ -1,7 +1,7 @@
 <?php
 /**
  * @package    hubzero-cms
- * @copyright  Copyright (c) 2005-2020 The Regents of the University of California.
+ * @copyright  Copyright 2005-2019 HUBzero Foundation, LLC.
```
hubzero-admin, line 4: Update this.

PR #1424, opened 2020-05-29
Add one more location to look at and copy the attachements from when …

ishunko:
…forking


ishunko:
Fixes https://qubeshub.org/support/ticket/1616

PR #1437, opened 2020-08-04
Com_installer Enhancements

erichhuebner:


datafatmunger:
From Slack:

jbg  5:55 PM
this code looks sort of dubious to me: https://github.com/hubzero/hubzero-cms/blob/6a515b7e1b2c6179e269e51b9d2c5f0d6ce41e5c/core/components/com_installer/admin/controllers/manage.php
publishTask, around line 237 ... wasn't part of the PR, i think ... dunno how i saw it with anthony's uber-scroll ... also maybe i don't understand the context. :shrug:

GitHubGitHub
hubzero/hubzero-cms
Platform for Scientific Collaboration. Contribute to hubzero/hubzero-cms development by creating an account on GitHub.


Jeff Kayser  5:59 PM
exit — Output a message and terminate the current script
5:59
That seems bad.
5:59
The rest is lost in the abstractness of the hub’s layers for me, so nothing jumps out in particular.

jbg  6:00 PM
i think you can successfully install 1 or more extensions, have a fail ... and falsely report success?
6:01
this success++ thing with if ($success)
6:01
combined with the continue in the fail?

Jeff Kayser  6:01 PM
Oh my, you’re right: $model = Extension::oneOrFail($id);
6:02
That [list of IDs] comes from a user-supplied var cid. (edited)

PR #1441, opened 2020-09-03
gitignore updates

erichhuebner:


erichhuebner:
Why not ignore /app/* ?

erichhuebner:
The current CMS GIT repo creates and tracks required directories in /app/*.  Ignoring the content of those directories was needed to ensure the required directories exist.

erichhuebner:
Trello card created for redesign of how we handle the creation of the needed directories in /app/

PR #1444, opened 2020-09-12
Commented out unused or incorrect mysql password locations

erichhuebner:


erichhuebner:
Muse migrations don't run correctly when finding incorrect mysql passwords.

nkissebe:
The code was fine, the password was stored wrong. Possibly changing the order of precedence would be a good idea. Pascal and I are going to normalize the mysql root user credential use and storage across hubs and create the /etc/mysql/hubmaint.cnf with credentials for muse to use.

What would be nice here is if there was an error message output to let the user know that the command failed and why (and in this case what file the failed credentials were read from).

erichhuebner:
Will be solved another way.  Closing.

PR #1446, opened 2020-09-17
Adding migrations for jos_auth_link and jos_auth_domain to speed up s…

damionjunk:
…tat queries.

### DIFF
--- a/core/migrations/Migration20200917000000AuthLinkIndex.php
+++ b/core/migrations/Migration20200917000000AuthLinkIndex.php
@@ -5,34 +5,37 @@ use Hubzero\Content\Migration\Base;
 // no direct access
 defined('_HZEXEC_') or die();
 
+/**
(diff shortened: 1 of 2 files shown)


core/migrations/Migration20200917000000AuthLinkIndex.php
```diff
@@ -0,0 +1,41 @@
...
+defined('_HZEXEC_') or die();
+
+/**
+ * Migration script for adding indexes to the #__auth_link table.
+ **/
+class Migration202000171000000AuthLinkIndex extends Base
+{
+	public function up()
+	{
+		if ($this->db->tableExists('#__auth_link'))
```
damionjunk, line 15: Only Run IF the index isn't already there.

damionjunk:
Conditionally run the add/drop's.

PR #1449, opened 2020-10-06
Retrieve previous months stats if users null

anthonyfuentes:
fixes: https://nanohub.org/support/ticket/363102


anthonyfuentes:
Also required updating nanoHUB's `app/plugins/resources/usagenewdata`

Committed on dev.nanohub.org (72fe23bdfcea01a6f0ff215849897e73485115fc)

PR #1450, opened 2020-10-07
Remove failing form token check

anthonyfuentes:
fixes: https://nanohub.org/support/ticket/378049


ishunko:
Need more research before merging.

damionjunk:
Adding trello card to pursue underlying fix asap, it is currently resolved like this on Nanohub, but this PR is generally being rejected.

PR #1453, opened 2020-10-08
[fix] Fix new modules under strict mode DB

zooley:
This mostly addresses an issue where creation of a new module would throw an under when using a database in strict mode. An empty string was being passed where strict mode expects the ID to be an integer or `null`.

PR includes some minor style and optimizations.


damionjunk:
@erichhuebner will reach out and discuss for more information.

erichhuebner:
After further review, this looks good.

PR #1469, opened 2020-11-06
Keep sorting query string when deleting a discussion

ishunko:


ishunko:
Fixes https://nanohub.org/support/ticket/373539

PR #1484, opened 2021-01-06
Update youtube url examples to https

ishunko:


ishunko:
as per https://mygeohub.org/support/ticket/1810

Issue #1489, opened 2021-02-20
Incorrect table referenced in AddTemplateEntry macro


mdlama:
I think, at the line listed below, the `insert` command should be referencing the table `#__template_styles`.

https://github.com/hubzero/framework/blob/c4f6010b263e660a54f938f41be295ee41a3bcf9/src/Content/Migration/Macros/AddTemplateEntry.php#L106

I got the following error when doing a clean `vagrant up` and the migrations were run:

```
vagrant@scotchbox:/var/www/public$ php muse migration --file=Migration20180217000000TplBmc.php -d=up -f
List all: all found files will be listed!
Warning: onBeforeMigrate hook 'UpdateTimezoneDatabase.php' resulted in an error: Insufficient privileges to update timezone tables.
Added extension entry for template "bmc"
Disabling "home" for all other templates (client "0")
Error: running up() resulted in

SQLSTATE[42S22]: Column not found: 1054 Unknown column 'template' in 'field list'

in app/templates/bmc/migrations/Migration20180217000000TplBmc.php
Migration failed! See log messages for details.
```

This was the command that led to the error:

```
$this->addTemplateEntry('bmc', 'bmc', 0, 1, 1, null, 0);
```

I looked at the code prior to the Macro-isization and noticed the table was `#__template_styles`:

https://github.com/hubzero/framework/commit/2393181d66f04bfd0535405a0cb778ac1d525132#diff-1304f94a6031037a97f7665d9c282502ce76ed4ce8b3f6d66748909f590678b8L1069

PR #1491, opened 2021-02-23
Fix: Update correct table when adding a template

mdlama:
Fixes: https://github.com/hubzero/hubzero-cms/issues/1489


mdlama:
Here's the qubeshub/hubzero-cms commit that addresses this (not sure why I didn't pull that one over!):

https://github.com/qubeshub/hubzero-cms/commit/ebd4b4302479fa85d37c988ea5d620710ad55d3b

PR #1522, opened 2021-05-27
Remove JS that does not let links in the toolbar to work

ishunko:


core/templates/kameleon/js/index.js
```diff
@@ -165,7 +165,6 @@ jQuery(document).ready(function($){
 	// Mobile device fix
 	$('#toolbar ul').on('click', function(e){
 		$(this).toggleClass('active');
-		e.preventDefault();
```
ishunko, line 168: The knowledge of why this was added on the first place was lost long time ago. According to the modern wisdom it is not needed.

PR #1524, opened 2021-06-29
Strengthen input validation of component names

pascal-meunier:
The path used with require_once may be invalid or manipulated with dots due to missing or incorrect input validation,  or the file may be unreadable.

### DIFF
--- a/core/libraries/Hubzero/Component/Loader.php
+++ b/core/libraries/Hubzero/Component/Loader.php
@@ -62,6 +62,7 @@ class Loader
 	 * @return  boolean
 	 */
 	public function isEnabled($option, $strict = false)
+	// Consider changing the default to true to prevent using component names that haven't been registered in the db and are therefore invalid
 	{
 		$result = $this->load($option, $strict);
 
@@ -126,8 +127,11 @@ class Loader
 		{
 			$option = implode('', $option);
 		}
-		$option = preg_replace('/[^A-Z0-9_\.-]/i', '', $option);
-		if (substr($option, 0, strlen('com_')) != 'com_')
+		// do not allow dots in component name to avoid directory traversal issues
+		$option = preg_replace('/[^A-Z0-9_-]/i', '', $option);
+		// prepend com_ to the name if it doesn't start with com_
+		// if option became empty due to the filtering, return an empty string
+		if ((strlen($option) > 0) && (substr($option, 0, strlen('com_')) != 'com_'))
 		{
 			$option = 'com_' . $option;
 		}
@@ -155,13 +159,13 @@ class Loader
 			$lang->load('tpl_' . $template, $this->app['template']->path, null, false, true);
 		}
 
+		$option = $this->canonical($option);
+
 		if (empty($option))
 		{
 			// Throw 404 if no component
 			$this->app->abort(404, $lang->translate('JLIB_APPLICATION_ERROR_COMPONENT_NOT_FOUND'));
 		}
-
-		$option = $this->canonical($option);
 
 		// Record the scope
 		$scope = $this->app->has('scope') ? $this->app->get('scope') : null;
--- a/core/libraries/Hubzero/Component/Loader.php
+++ b/core/libraries/Hubzero/Component/Loader.php
@@ -129,11 +129,21 @@ class Loader
 		}
 		// do not allow dots in component name to avoid directory traversal issues
 		$option = preg_replace('/[^A-Z0-9_-]/i', '', $option);
-		// prepend com_ to the name if it doesn't start with com_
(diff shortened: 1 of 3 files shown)


zooley:
"... or manipulated with dots due to missing or incorrect input validation ..."

The path is derived from the model's name which is either a set property of the controller or derived by singularizing the controller's name and, thus, no external inputs are accepted or should even involved in the process.

pascal-meunier:
Hi Shawn, the API infers the component from the URL which results in an observable and demonstrated exception when the URL contains dots at the right place, as it tries to open the component "com_..".

> The path is derived from the model's name which is either a set property of the controller or derived by singularizing the controller's name and, thus, no external inputs are accepted or should even involved in the process.

pascal-meunier:
In case I'm not being clear, this patch fixes an exception triggered by the API when the URL contains '..' where the component name is expected.

zooley:
That's a problem that should be handled via the component loader. That is, the check shouldn't be needed as the system shouldn't ever allow it to reach that point. The real issue seems to be here:

https://github.com/hubzero/hubzero-cms/blob/77136aa43a7d9506504acea8ba16b7b5613a2468/core/libraries/Hubzero/Component/Loader.php#L129

I can't think of any reason why periods would be allowed in component names. Probably not even dashes and should just be `[^A-Z0-9_]` but that's less an issue. Then check if the resulting option is just 'com_':

```
	public function canonical($option)
	{
		if (is_array($option))
		{
			$option = implode('', $option);
		}
		$option = preg_replace('/[^A-Z0-9_-]/i', '', $option);
		if (substr($option, 0, strlen('com_')) != 'com_')
		{
			$option = 'com_' . $option;
		}
		if ($option == 'com_')
		{
			$option = '';
		}
		return $option;
	}
```

There's a secondary issue that these lines:

https://github.com/hubzero/hubzero-cms/blob/77136aa43a7d9506504acea8ba16b7b5613a2468/core/libraries/Hubzero/Component/Loader.php#L164

https://github.com/hubzero/hubzero-cms/blob/77136aa43a7d9506504acea8ba16b7b5613a2468/core/libraries/Hubzero/Api/Component/Loader.php#L34

Need to be moved up to right before the `if (empty($option))` checks.

```
		$option = $this->canonical($option);

		if (empty($option))
		{
```

That should guard against invalid component names. Then it should fall through to the "is enabled" checks. There is a flag for strict checking (i.e., if it isn't in the database and enabled, throw a 404) but the component loader isn't checking with that flag purely because it was easier to develop without it (drop a component in place with no DB entry and it still loads). That may simply have to be enforced going forward.

`if ($this->isEnabled($option))` should probably be `if ($this->isEnabled($option, true))`

pascal-meunier:
Thank you Shawn, I made the changes you suggested instead of my initial ones.  I'll keep the idea of the strict checking for when we are ready to verify it won't break anything in production hubs.

core/libraries/Hubzero/Component/Loader.php
```diff
@@ -126,8 +127,11 @@ public function canonical($option)
 		{
 			$option = implode('', $option);
 		}
-		$option = preg_replace('/[^A-Z0-9_\.-]/i', '', $option);
-		if (substr($option, 0, strlen('com_')) != 'com_')
+		// do not allow dots in component name to avoid directory traversal issues
+		$option = preg_replace('/[^A-Z0-9_-]/i', '', $option);
+		// prepend com_ to the name if it doesn't start with com_
+		// if option became empty due to the filtering, return an empty string
+		if ((strlen($option) > 0) && (substr($option, 0, strlen('com_')) != 'com_'))
```
zooley: You may still want to add the following after this `if` state:

```
if ($option == 'com_')
{
	$option = '';
}
```

If I call `{host}/api/index.php?option=com_..`, the periods will be filtered and the resulting `$option` is just "com_" which will pass the `if (empty($option))` check.

pascal-meunier:
Would it be wrong to abort inside that function and so reduce complexity and save resources, as in:
```
if ((strlen($option) == 0) || ($option == 'com_')) $this->app->abort(404, $lang->translate('JLIB_APPLICATION_ERROR_COMPONENT_NOT_FOUND'));
if (substr($option, 0, strlen('com_')) != 'com_')
{
	$option = 'com_' . $option;
}
```

zooley:
I wouldn't. That method is used in a variety of places to sanitize `$option` values and there may be cases where some piece of code is doing checks or whatever and you don't want a 404 getting thrown in the middle of it. A bit of a separation of tasks issue. The `canonical` method's sole purpose is to sanitize and normalize a given string.

pascal-meunier:
Thank you Shawn.

PR #1530, opened 2021-09-02
Pub batch import

mdlama:
Consists of 4 commits that extend publication batch import, with a couple of fixes thrown in. Here's my attempt at a summary of changes:

- New select input for choosing master type
- Can now include arbitrary metadata (goes in the `metadata` field in the publication versions database)
- Modified the code so batch importing can be called via an api (didn't end up using this but left it in as it might be useful in the future; could also be problematic as I removed a security token)
- Added the ability to import DOIs
- Added the ability to specify which gallery image should be used as the thumbnail
- Fixed a bug with importing category information

I've attached the test data that I used, which includes the xml that I imported.

[Resource_1598.zip](https://github.com/hubzero/hubzero-cms/files/7095719/Resource_1598.zip)

You can see the updated schema in the commits.


mdlama:
@erichhuebner Bringing your attention to this one

nkissebe:
aa27c0... I'd delete rather than comment out the vcs hash reference.  (line 650/661) There is a "use Components\Publications\Tables;" in the file so shouldn't need to change the class reference to include the whole thing. Minor thing, just prefer to either have the use statement or use the complete class path not have both.

9337e83... I would really rather not make 'api' modes of regular calls. We have an api layer, this should be refactored to be callable from both an api endpoint and the site endpoint (if needed).

Not saying this PR should be blocked, just my input.

nkissebe:
Re: api.... and failing putting it in the api layer I think we have an existing practice on how to do it... lets see if I can find it...

nkissebe:
&no_html=1 , I think you will even find its use in com_publications  (I may be wrong about its use, but I think thats the purpose)

PR #1532, opened 2021-09-02
[fix] Added check for local DOI before updating

mdlama:
Added this to check if batch imported doi has a different namespace/shoulder from ones minted by the hub. If different, don't try and auto-update metadata (was throwing an error).


mdlama:
@erichhuebner ...and this one

nkissebe:
It'll work as long as shoulder never contains a special regular expression character

PR #1535, opened 2021-09-10
Properly handle user agreement checkbox on the registration form

ishunko:


erichhuebner:
Please check that this solution also works with the modal pop-up in the members profile when registration is set to require TOU acceptance on next login.

ishunko:
> Please check that this solution also works with the modal pop-up in the members profile when registration is set to require TOU acceptance on next login.

Checked. Everything works fine.

PR #1549, opened 2022-02-14
Add parentheses to divide out units prior to function call

jsperhac:
Ran into an issue compiling the lucent (https://gitlab.hubzero.org/hubzero/lucent-the-minimal-hubzero-template) template with `less`. The offending location is in the core file variables.less.

*Error message:* 
`ArgumentError: Error evaluating function `percentage`: argument must be a number in /core/assets/less/variables.less on line 337, column 27`

Same error was thrown on line 338, which performs same operation on related dimensions.

*`lessc` version*
used for compile:
lessc 4.1.2 (Less Compiler) [JavaScript]

*Fix:*
Add nested parentheses to indicate the division step should be performed prior to function call to percentage(). This ensures that the units are divided out of the argument passed to the function. The resulting argument is numeric and its percentage can be computed.

*Test:*
Verified that compilation of the less files into css proceeds without errors once this single change is made to lines 337 and 338

### DIFF
--- a/core/assets/less/variables.less
+++ b/core/assets/less/variables.less
@@ -327,8 +327,8 @@
 
 // Fluid grid
 // -------------------------
-@fluidGridColumnWidth:    percentage(@gridColumnWidth/@gridRowWidth);
-@fluidGridGutterWidth:    percentage(@gridGutterWidth/@gridRowWidth);
+@fluidGridColumnWidth:    percentage((@gridColumnWidth/@gridRowWidth));
+@fluidGridGutterWidth:    percentage((@gridGutterWidth/@gridRowWidth));
 
 // Breaking the grid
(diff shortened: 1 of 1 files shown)


erichhuebner:
Further testing to be done before merging into the dev branch.

jsperhac:
Confirmed that the 'less' change proposed here compiles against a basic template successfully with either less v3 (v 3.11.3) and less v4 (4.1.2). However, the current dev branch (without this change) compiles against the same template with less v3 only. 

Thanks to @nkissebe for pointing out that: "the new version of less has different default for division," according to  https://lesscss.org/usage/#less-options-math

Note that the compilation of the basic hubzero template (lucent) yielded a deployable css file. This looks safe to merge.

PR #1550, opened 2022-02-22
feat(csp): Content Security Policy plugin

zooley:
This is a basic system plugin for setting the content security policy
headers for a hub.

This was initially started in May 2019 but wasn't finished until now.
Contributing due to expressed interest:

https://help.hubzero.org/support/ticket/12697

Refs: https://developer.mozilla.org/en-US/docs/Web/HTTP/CSP


zooley:
Reverted accessibility commit as it wasn't meant for this PR. It will be a separate one later on.

PR #1552, opened 2022-03-16
[fix] Offset error with double slash in URL

mdlama:
Fixes: https://qubeshub.org/support/ticket/4791


nkissebe:
From ticket. Just including for clearer documentation on what this change is doing and why:

Changing that to "$segments = array_values(array_filter($segments));" seems to fix the issue (i.e. re-index since the first index with a \\ is 1 instead of 0, hence the offset error). I'll put in a pull request with the fix.

PR #1555, opened 2022-04-25
[fix] Tags for versions where being deleted

mdlama:
Legacy code removed from when tags were associated with publications
  and not versions. This code was inadvertently deleting tags for
  versions with ids matching the removed publication id.


mdlama:
Wait, this pull request might be incomplete. Checking code now to make sure version tags are being deleted. Will add additional commits soon if needed.

mdlama:
Okay, looks like version tags are deleted appropriately. I think this is the only code change needed to fix the issue.

PR #1560, opened 2022-06-02
Fix the tools editing UI issue on larger screens

ishunko:
Solves the issue mentioned on
https://gateway.delta.ncsa.illinois.edu/support/ticket/6
https://geodynamics.org/support/ticket/19
https://vhub.org/support/ticket/2571


dbenham:
@dbenham - Needs deployment. Hotfix it for now.

PR #1568, opened 2022-07-26
Fix for NanoHub Ticket #378563

JackS9:
This adds a help entry for adding an external link to a course outline. Fixes NanoHub Ticket #378563.

### DIFF
--- a/core/components/com_courses/site/help/en-GB/builder.phtml
+++ b/core/components/com_courses/site/help/en-GB/builder.phtml
@@ -24,11 +24,11 @@ defined('_HZEXEC_') or die();
 	frameborder="0" allowfullscreen></iframe>'); ?>
 </pre>
 
-<a name="link"></a><h2>Embedding an exteral link into your outline</h2>
+<a id="link"></a><h2>Embedding an exteral link into your outline</h2>
 <p>
 	Embedding an external link into your course outline is simple. Just enter the URL in the text box provided, for example:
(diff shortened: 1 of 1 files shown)


nkissebe:
Jack, can you change the example to use a URL like example.com

core/components/com_courses/site/help/en-GB/builder.phtml
```diff
@@ -22,4 +22,13 @@ defined('_HZEXEC_') or die();
 	<?php echo htmlspecialchars('<iframe width="560" height="315" 
 	src="http://www.youtube.com/embed/tW1-fSRiAdc?rel=0" 
 	frameborder="0" allowfullscreen></iframe>'); ?>
-</pre>
\ No newline at end of file
+</pre>
+
+<a name="link"></a><h2>Embedding an exteral link into your outline</h2>
```
zooley: Minor note: The `name` attribute on anchors isn't valid in HTML 5. The preferred method is to use an `id`.

```
<h2 id="link"> ...
```

There's undoubtedly a number of these throughout the core that need to get updated.

dbenham (commented):
I changed name to id, and used a generic example.com link in the example.

dbenham:
merging change to dev, this is not a critical fix, so no hot fix will be created. But going to review code deployment procedures, not specifically for this fix, but fixes in general.

PR #1575, opened 2022-09-15
Remove special symbols escape to properly display quotes.

ishunko:
Order quotes by date, newest first


nkissebe:
I am concerned about the removal of the escape. Is it possible the data was stored already escaped and shouldn't have been? I'm not super familiar with this.

PR #1577, opened 2022-09-30
Update Migration20150202134839PlgAntispam.php

dbenham:
fix from dev made by Erich, commit c0b188c9fdcfab118b14abfe22f63e0e494ec3af.

Nick, if there is a better way to get this fix into 2.2.22, please let me know.


nkissebe:
This commit doesn't really match the one you reference. It still includes the original addPlugin call for blacklist.

nkissebe:
I cherry picked the referenced commit to 2.2.22 for what its worth

PR #1585, opened 2022-11-04
Exclude Pending resources from Solr search result set

jsperhac:
## Summary

This PR fixes a defect reported by Nanohub users that allowed Pending (unpublished) resources to be returned in searches of the site and accessed by unprivileged users. The defect was a result of several issues: incorrect function calls, and failure to check all relevant resource statuses.

## Motivation

Nanohub ticket #412067 

## Testing

Solution was tested on a Hubzero VM on which various public, private, and protected resources were created in various publication states. Fix was quietly added to Nanohub in September 2022.


nkissebe:
Looks like e131f0b7162e60af706d6b0b4ba38fab421918d7 from 2019 already had these, nanohub was just behind

PR #1586, opened 2022-11-04
Fix Solr full indexing option in administrator cron component

jsperhac:
## Summary

This PR makes two small fixes to the Hub's administrator site cron component for Solr indexing. 

The component is only available for use once it is enabled in the Components:Cron admin interface, and when unused, these issues escaped detection.

## Motivation

Once the component was enabled, attempting to reindex the Hub for the Solr search engine threw two errors: 

1. Unable to find Class 'Components\Search\Models\Solr\SearchComponent'
2. Missing argument 1 for Components\Search\Models\Solr\SearchComponent::indexSearchResults()

This fix corrects both of those issues.

## Testing

This fix was verified on a Hubzero VM and on Nanohub.


nkissebe:
I manually applied this in your name. I wasn't sure how the merge commit would appear

PR #1593, opened 2023-01-24
Add BrowserStack text to README

jsperhac:
Citing BrowserStack enables them to grant us access to their testing utilities: https://www.browserstack.com/open-source


dbenham (approved):
lgtm

PR #1594, opened 2023-01-24
fix parsing bug for /content/article

JackS9:
This fixes bugs in parsing hub/content/article/<cat-slug>/<article-slug> and hub/content/categories.

### DIFF
--- a/core/bootstrap/Site/routes.php
+++ b/core/bootstrap/Site/routes.php
@@ -586,18 +586,42 @@ $router->rules('parse')->append('content', function ($uri)
 	{
 		// Build the path
 		$path = array();
+		$isCategory = false;
 		for ($i=0; $i < ($count-1); $i++)
 		{
+			// Skip "/content/article" prefix in path if present
+			if ($segments[$i] == 'content' || $segments[$i] == 'article')
+			{
+				continue;
+			}
+			if ($segments[$i] == 'category')
+			{
+				$isCategory = true;
+				continue;
+			}
 			$path[] = $segments[$i];
 		}
 
+		if ($isCategory)
+		{
+			$path[] = $segments[$count-1];
+		}
 		$path = implode('/', $path);
 
-		// Now, do query (path is all but last segment, and last segment is article alias)
-		$query  = "SELECT con.`id` FROM `#__content` AS con";
-		$query .= " LEFT JOIN `#__categories` AS cat ON con.catid = cat.id";
-		$query .= " WHERE con.state=1 AND con.`alias` = " . $db->quote(strtolower($segments[$count-1]));
-		$query .= " AND cat.`path` = " . $db->quote(strtolower($path));
+		if ($isCategory)
+		{
+			// Do category query using path as category
+			$query  = "SELECT cat.`id` FROM `#__categories` AS cat";
+			$query .= " WHERE cat.`extension`='com_content' AND cat.`path` = " . $db->quote(strtolower($path));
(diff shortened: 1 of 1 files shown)


dbenham:
@nkissebe, can you review this? I'd like a second set of eyes on this one. Thx.

dbenham:
@JackS9 

Nick and I were a bit concerned a bit this code is so old and made it this far without being noticed before and that we might be looking at the possibly the end users (or even us) are assuming this 'feature' might work differently than intended, and that it actually was working correctly.

Would you mind doing a brief summary about this so we could review this in a bit more detail?

Thanks.

JackS9:
Several article-related modules (mod_articles_category, mod_articles_latest, mod_articles_news, mod_articles_popular, mod_latest, mod_related) use helper() to call Route::url() with \Components\Content\Site\Helpers\Route::getArticleRoute() to generate article links.  These generated links have the pattern /content/article/_cat-alias_/_article-alias_.  However, these paths are not properly parsed.  The ‘/content/article’ prefix turns out to be unnecessary.  

The path building process follows:

Route::url() —> /core/libraries/Hubzero/Routing/Router.php (via /core/libraries/Hubzero/Facades/Route.php)
  —> build() 
    —> createUri()
      Loop over “build” rules pre-registered by /core/bootstrap/Site/Providers/RouterServiceProvider.php via /core/libraries/Hubzero/Routing/Manager.php
        —> /core/bootstrap/Site/routes.php (for Component = ‘content’)
          —> /core/components/com_content/site/router.php->build()

The parsing process similarly ends up at /core/components/com_content/site/router.php->parse(), which in turns ends up looping over the pre-registered “parse” rules in core/bootstrap/Site/routes.php (for Component = ‘content’)

One can either try to fix the path building process to not add the ‘/content/article’ prefix, or fix the parsing process to properly handle the ‘/content/article’ prefix. I chose the latter, which does not remove any functionality and will still handle previously generated links, which have since broken.

Likewise /content/categories generates paths of the pattern /content/category/_cat-alias_, which also are not properly parsed by core/bootstrap/Site/routes.php (also for Component = ‘content’).  So I extended the fix to the above to also handle the categories case.

JackS9:
Can this be temporarily put in /app/bootstrap/site/routes.php on sites that need a fix until we find a better solution?

dbenham:
@JackS9, yeah, where do you want me to put it?

JackS9:
I just tried moving the modified version of routes.php from /core/bootstrap/Site to /app/bootstrap/site on stage.plantingscience.org, and it did not override the core version.  I then noticed that it's 'site' on /app/bootstrap, not 'Site' as it is on /core/bootstrap, so I renamed 'site' to 'Site' on /app/bootstrap to see if that made a difference, but it still was not overriding /core/bootstrap.  So I'm still not sure how to override a bootstrap routine without patching core.  Any ideas?

dbenham:
Checking with Nick to see if we can override anything in bootstrap, if not, I'll deploy this to core on stage.plantingscience.org, we can check it out, and then approve this merge to core.

JackS9:
This is already on PlantSci.

JackS9:
I now have a fix that only involves the parser in core/components/com_content/site/router.php. It just looks up the id using the alias if the id (alias) is not an integer.  I'll commit it (as branch route-bug2) and submit a PR shortly.

JackS9:
Closing this PR since another work-around has been submitted.

PR #1606, opened 2023-02-28
PLANTSCI-29 500 Error in forum for new access groups

JackS9:
Adds default forum access level (other) and corresponding color_access for undefined values (>5), with same access and color as 'public'.


dbenham:
@JackS9 

Do values higher than 5, do those actually correspond to public groups? How many are currently undefined?

JackS9:
Unfortunately, these routines only handle standard access levels (1-5), not custom (site-specific) levels (>5) that are defined under Members > Access > Viewing Levels. The 500 error was created by color_access not being defined (not having a default value), so I was just setting it to ‘public’. It’s only used to build a classname used in the CSS, not to set any actual access level.  The actual access level (a numeric value) is passed through unchanged. I also created a label ‘Other’ to identify them as non-standard (not having access to the label in this context).

To make it more adaptable to custom access levels would require much more work, including cross-table lookups to translate ids to labels and maybe adding columns for color, etc.

- Jack

JackS9:
Actually, these changes only affect styling (CSS).  All access (viewing) control is handled elsewhere.

nkissebe:
I don't know about forums, but "access levels" aren't the same everywhere in the system For instance in resources they are just the hard coded public, registered, special, private. I'm not sure why there is an 'other' and what its used for. I am a little concerned about where this is coming from. It shouldn't be getting a value outside that range I don't think

dbenham:
Isn't the first fix to supress an error based on data/configuration/code we can't explain. I hate it, but don't see any other way forward.

So the value displayed for these groups on screen says 'other' and it has the same color as label for the public groups?

Only change I would like is to have a unique color set to the 'other' groups. 

$color_access = 'public'; <----- does this code just apply a style of some kind then?

nkissebe:
Well we should probably see where the value is coming from, its likely going to cause other issues.

JackS9:
These custom access (viewing) levels are defined on the admin side via Members > Access > Viewing Levels. They are used in other places to control access, but here it’s only used to help set classes, like color_access for CSS styling on the admin pages. If the switch statement doesn’t include a case for that level, it has no default and gives a 500 error.

JackS9:
From support ticket: https://plantingscience.org/support/ticket/275
———-
Sad to report that the 'Forum' component screen in the admin portal has broken for some reason. It's giving us a 500 error that says, "Undefined variable: color_access"
This occurred after we changed the Access group on a forum category in our F2 Professional Development group. We're in the setup/testing phase there, so this isn't a crisis, and fortunately the front side forums still seem to be functioning normally.
Here's a video showing what I see: https://watch.screencastify.com/v/Qp0Qs1W3HWhNaj94j6Ru

JackS9:
Would it help to define a new color (in CSS) for class 'other' instead of re-using 'public'? Dynamically defining new colors for each custom access (viewing) level would be non-trivial.

The colors for .access.public, access.private,,,, are defined in https://stage.plantingscience.org/core/templates/kameleon/css/index.css.  I can just add a new color for .access.other.

nkissebe:
What you have is fine. Its the same shawn did in the categories view.
This can be merged as far as i am concerned.

JackS9:
I pushed a new commit to the plantsci29 branch with 'other' now being a valid value for color_access in the template CSS.  I also made the changes more consistent across categories, sections and threads.  Again, these changes only affect styling/labelling in the admin pages and have no actual effect on access.

JackS9:
Should I withdraw this one and resubmit using the later commit?

JackS9:
I just realized that I edited the index.css file directly instead of the index.less file.  I've now edited the index.less file, but I don't know how one properly make changes to a .less file and get it compiled to a .css files.

dbenham:
If you push another commit to this PR that is fine. and I see you already did that.

dbenham:
LGTM

dbenham (approved):
LGTM

JackS9:
Awesome.  How do we get the index.less file compiled and replace the index.css file?

PR #1607, opened 2023-03-05
alternate approach to fixing article/category route parsing bug

JackS9:
This PR supercedes PR 1594 to fix parsing issues with routes like content/article/category-alias/article-alias or content/category/category-alias, which don't have numerical item ids.  This approach only involves the parser within the Content component, not the parsing rules generated during bootstrap.

### DIFF
--- a/core/components/com_content/site/router.php
+++ b/core/components/com_content/site/router.php
@@ -286,8 +286,26 @@ class Router extends Base
 		if (!isset($item))
 		{
 			$vars['view'] = $segments[0];
-			$vars['id']   = $segments[$count - 1];
-
+			$id = $segments[$count - 1];
+			if ($id+0 == 0) // check if an alias, not an integer if
+			{
+				if($vars['view'] == 'article')
+				{
+					$query = 'SELECT id FROM `#__content` WHERE alias = ' . $db->Quote($id);
+				}
+				else if($vars['view'] == 'category')
+				{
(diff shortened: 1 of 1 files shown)


dbenham:
Jack, can you link the ticket this is supposed to be a fix for?

JackS9:
It dates back to PSC 77, but the issue has appeared in multiple places since. See, for example, PLANTSCI 269 and 271.

dbenham:
@nkissebe - this is one of those core changes you wanted to be consulted on ;) Can you please take a look at it and let me know what you think? I hate changes to such old code, always second guessing myself that the bug lived for this long before being discovered ;)

jessewoo:
@JackS9 - I'm adding this to stage.plantingscience.org

core/components/com_content/site/router.php
```diff
@@ -286,8 +286,26 @@ public function parse(&$segments)
 		if (!isset($item))
 		{
 			$vars['view'] = $segments[0];
-			$vars['id']   = $segments[$count - 1];
-
+			$id = $segments[$count - 1];
```
jessewoo, line 289: @JackS9 - what is this supposed to do? I don't understand the context of the fix.

JackS9: This particular snippet is to delay the $vars['id'] assignment until after other logic (following this snippet) is processed, mainly to check if it's numeric (id) or an alias, in which case the id is look up in the database.

JackS9:
This PR is being replaced by PR 1624.

PR #1608, opened 2023-03-15
Purr admin tags displaying improperly

dbenham:


core/components/com_publications/helpers/tags.php
```diff
@@ -351,7 +351,7 @@ public function count_tags($admin=0)
 	public function get_tag_cloud($showsizes=0, $admin=0, $objectid=null)
 	{
 		$cloud = new Cloud($objectid, $this->_tbl);
-		return $cloud->render();
+		return $cloud->render('html', array('admin' => $admin));
```
dbenham, line 354: the render function needed a filter array passed in to remove admin tags

core/components/com_publications/site/views/about/tmpl/default.php
```diff
@@ -216,7 +216,7 @@
 	</div>
 <?php } ?>
 <?php if ($this->publication->params->get('show_tags')) {
-	$this->publication->getTagCloud( $this->authorized );
+	$this->publication->getTagCloud(User::authorise('core.admin')?1:0);
```
dbenham, line 219: $this->authorized wasn't the best way to determine if admin tags should be shown, public users ended up having this set to true, which was the core of this bug.

core/components/com_tags/models/cloud.php
```diff
@@ -231,10 +231,7 @@ public function tags($rtrn='', $filters=array(), $clear=false)
 				// Filter out admin tags - JBG
 				$results->whereNotIn('admin', [1]);
 			}
-			else
-			{
-				$results->whereEquals('admin', (int) $filters['admin']);
-			}
+
```
dbenham, line 234: I have no idea why this else statement was ever included, it basically says to include only admin tags if the admin filter was not set, I couldn't discover a way to even trigger this code.

dbenham (commented):
self reviewed, tested on dev.purr.purdue.edu

dbenham:
PURR-87/PURR ticket 2442 - admin tags showing up for non admin users

PR #1613, opened 2023-04-18
Date Fields on Articles Behaving Strangely During Copy

jessewoo:
- https://sdx-sdsc.atlassian.net/browse/PLANTSCI-39
- https://plantingscience.org/support/ticket/286

Upon copying an existing article, the create date of the new article is wrong and not the current date. When user changes the start publishing and finish publishing date, the date jumps a few hours back from what was selected on the calendar icon. 

**Brief summary of the fix**
- Upon the copy of the original article, it will be saved with current date for the creation and publishing up date, the alias will also be unique
- The start publishing and finish publishing date will be set at UTC since the calendar selection is in UTC as well. The standardization from all the Hubzero dev will be UTC. 

Manual testing

No

Yes - Nick and David

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,0 +1,10 @@
+# Before you submit this pull request, please make sure:
+
+- [ ] Make sure there are links in the PR to the source JIRA card(s) and hubzero ticket(s)
+- [ ] Include a brief summary of the issue *in your own words*
+- [ ] Include a breif summary of the fix
+- [ ] Include a brief summary of your testing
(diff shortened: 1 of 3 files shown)


dbenham:
@Jesse, I think you might not have checked in the files with your fix. The two modified files I currently see in this PR seem related to your fix.

dbenham (commented):
I'm not sure you have the right files included for this PR. Can you check and make sure your changes are included here? 

You should just be able to re-push any missing code to this branch and this PR should reflect the most recent version of the files you pushed.

jessewoo:
issues with this pull request.

PR #1615, opened 2023-04-20
Allow groups to show in tag search results for public visitors

jessewoo:
- https://sdx-sdsc.atlassian.net/browse/VHUB-9
- https://theghub.org/support/ticket/2601

When a user adds a tag to a group, and upon searching for a tag, that specific group won't appear in the search results for public visitors but will appear for registered and logged in users. 

An example is on the search results of https://theghub.org/tags/ghub. The ISMIP6 group has the associated "Ghub" tag. But that ISMIP6 group only shows up in the results when the user is registered and logged in. The issue is how we can make groups show up for specific tags in the search results for non-registered (public) users.

**Brief summary of the fix**
- The search SQL statement originally filtered out those groups, so removed the filter for public users. 

Manual testing

No

Yes - Nick and David

### DIFF
--- a/core/plugins/tags/groups/groups.php
+++ b/core/plugins/tags/groups/groups.php
@@ -61,6 +61,11 @@ class plgTagsGroups extends \Hubzero\Plugin\Plugin
 			$from = " JOIN #__xgroups_members AS m ON m.gidNumber=a.gidNumber AND m.uidNumber=" . (int)User::get('id', 0);
 		}
 
+		// Allow tag to be visible to public visitors, previously tag was visible only for registered and logged in users
+        if (User::isGuest()) {
+            $from = '';
+        }
+
 		// Build the query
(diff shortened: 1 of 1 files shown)


dbenham (approved):
LGTM.

Next time for the testing feedback, I'm looking for more of a description of what you did to test to confirm your change works as expected:

_For a bug fix, something like:_
"I ran the code before my fix with some input and it returned/behaved like X, after I applied my fix I re-ran the test with the same input and the expected result of (whatever expected restult) showed up properly"

**OR**

_for a new feature, something like_:
"This was a new feature, I ran tests with various inputs, range tested inputs 1,2,3,4, and made sure there were max length tests on fields 1,2,3,4. Logged out, feature was unavailable (this is the expected behavior). Talked with client, they took a look at it and said it was fine, or I went through the design doc (section xxx) to make sure all design criteria were properly"

PR #1623, opened 2023-05-23
[PLANTSCI-47] updates to groups-forum to support planting science with custom view …

JackS9:
- JIRA task: https://sdx-sdsc.atlassian.net/browse/PLANTSCI-47
- Update Groups Forum to handle custom view access levels to filter discussions
- Extended view access levels to handle custom levels as well as standard levels (public, registered, private) on both client and admin sides.
- Tested on stage.plantingscience.org by client
- Hotfix required?  Yes, to get changes in production for Planting Science hub ASAP

### DIFF
--- a/core/components/com_forum/admin/views/categories/tmpl/display.php
+++ b/core/components/com_forum/admin/views/categories/tmpl/display.php
@@ -32,6 +32,13 @@ if ($canDo->get('core.delete'))
 }
 Toolbar::spacer();
 Toolbar::help('categories');
+$db = App::get('db');
+$query = $db->getQuery()
+	->select('id')
+	->select('title')
+	->from('#__viewlevels');
+$db->setQuery($query->toString());
+$access_levels = $db->loadAssocList();
 ?>
 
 <form action="<?php echo Route::url('index.php?option=' . $this->option . '&controller=' . $this->controller); ?>" method="post" name="adminForm" id="adminForm">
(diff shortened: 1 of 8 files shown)


core/components/com_forum/admin/views/categories/tmpl/display.php
```diff
@@ -32,6 +32,13 @@
 }
 Toolbar::spacer();
 Toolbar::help('categories');
+$db = App::get('db');
+$query = $db->getQuery()
+	->select('id')
+	->select('title')
+	->from('#__viewlevels');
+$db->setQuery($query->toString());
+$access_levels = $db->loadAssocList();
```
nkissebe, line 41: This code should get moved to the controller and $access_levels passed into the view

nkissebe: same issue in the other views

JackS9: Good idea.  

BTW, there are several potential issues that need attention before pushing this beyond group forums on Planting Science, which would be a good topic for one of our dev sessions. 

Custom view access levels (beyond the default public, registered, private and special) are rarely checked for or leveraged with other assets (articles, forums, resources, publications,collections, blogs,..). Opening this up can lead to ambiguities when combined with group membership, project team membership, resource authorship, etc. Further hierarchical restrictions implied by categories, sections, threads, etc. can easily be overlooked.

nkissebe:
We are not accepting this PR for the moment and just leaving it on planting science. We are concerned that the incomplete implementation may cause more confusion than what it fixes.

PR #1624, opened 2023-05-23
[PLANTSCI-47] latest version of content router/parser

JackS9:
- JIRA Task: https://sdx-sdsc.atlassian.net/browse/PLANTSCI-47
- This update to the content router/parser addresses several issues related to parsing of SEF URLs for articles and categories generated elsewhere in HUBzero code using Route::url(), including a /content/article or /content/category prefix, aliases including dashes (also used internally for slugs in place of colons), and aliases beginning with numeric digits (being confused with id portion of a slug).
- The parser changes include: looking up aliases in the database and replacing with ids, skipping the conversion of dashes to colons for cases where it's clearly not a slug, performing more rigorous (regex) test to discern alias beginning with numeric digit from an id.
- These changes have been rigorously tested on stage.plantingscience.org.
- These changes have already been patched on plantingscience.org production site.
- **This PR supersedes PR 1607,** which is still pending, and will be withdrawn. 

**Thanks!**

### DIFF
--- a/core/components/com_content/site/router.php
+++ b/core/components/com_content/site/router.php
@@ -278,7 +278,10 @@ class Router extends Base
 
 		for ($i = 0; $i < $count; $i++)
 		{
-			$segments[$i] = preg_replace('/-/', ':', $segments[$i], 1);
+			if (is_numeric($segments[$i][0]))
+			{
+				$segments[$i] = preg_replace('/-/', ':', $segments[$i], 1);
+			}
 		}
 
 		// Standard routing for articles.  If we don't pick up an Itemid then we get the view from the segments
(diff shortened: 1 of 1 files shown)


core/components/com_content/site/router.php
```diff
@@ -278,15 +278,38 @@ public function parse(&$segments)
 
 		for ($i = 0; $i < $count; $i++)
 		{
-			$segments[$i] = preg_replace('/-/', ':', $segments[$i], 1);
+			if (is_numeric($segments[$i][0]))
+			{
+				$segments[$i] = preg_replace('/-/', ':', $segments[$i], 1);
```
nkissebe, line 283: Should the check here be if segment ends in -####, as -season2023 wouldn't be an article number we would have to consider later?

core/components/com_content/site/router.php
```diff
@@ -278,15 +278,38 @@ public function parse(&$segments)
...
 
 		// Standard routing for articles.  If we don't pick up an Itemid then we get the view from the segments
 		// the first segment is the view and the last segment is the id of the article or category.
 		if (!isset($item))
 		{
 			$vars['view'] = $segments[0];
-			$vars['id']   = $segments[$count - 1];
+			$id = $segments[$count - 1];
+			//if ($id+0 == 0) // check if an alias, not an integer id
+			if (preg_match('#[^0-9]#',$id))
```
nkissebe, line 294: An alias could have a digit in it, so not clear to me what this check is for

core/components/com_content/site/router.php
```diff
@@ -300,7 +323,15 @@ public function parse(&$segments)
 			if (strpos($segments[0], ':') === false)
 			{
 				$vars['view'] = 'article';
-				$vars['id']   = (int)$segments[0];
+				$id   = $segments[0];
+				//if ($id+0 == 0) // check if an alias, not an integer id
+				if (preg_match('#[^0-9]#',$id))
```
nkissebe, line 328: Same question as above

nkissebe:
Jack and I should have a zoom call and go over the original problem and the solution.

core/components/com_content/site/router.php
JackS9: I tried a couple ways to discern an id from an alias, and each seemed to have a flaw.  This one just checks for the presence of any non-digit in the string to decide it's not an id.  Any advice would be appreciated.  

[I probably shouldn't have used a non-standard regex delimiter (# instead of /) - an old carry-over from sed/awk days of parsing mathematical expressions.]

dbenham:
@nkissebe  - is this worth keeping?

dbenham:
Closing due to inactivity

PR #1625, opened 2023-05-29
[VHUB-25] propagate group alias (group_cn) to author step in creating new group…

JackS9:
- JIRA task: https://sdx-sdsc.atlassian.net/browse/VHUB-25
- Support ticket: https://theghub.org/support/ticket/2739
- Summary of issue: The group owner for a new group resource does not default to current group
- Summary of fix: The group alias (group_cn) is now propagated as a query parameter from step-to-step (compose through authors) to help set the default (selected) group in the drop down menu of groups in which the current user belongs
- Summary of testing: Tested on personal AWS development machine by creating a new group resource.  Current group is now selected by default where it was not before.  
- Hotfix required?  Possibly, if vhub/ghub (theghub.org) wishes to have it implemented before next rollout

### DIFF
--- a/core/components/com_resources/site/views/steps/tmpl/attach.php
+++ b/core/components/com_resources/site/views/steps/tmpl/attach.php
@@ -28,11 +28,13 @@ $this->css('create.css')
 
 <section class="main section">
 	<?php
+		$this->group_cn = Request::getString('group','');
 		$this->view('steps')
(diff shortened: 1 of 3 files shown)


dbenham (commented):
lgtm

dbenham (approved):

PR #1630, opened 2023-06-23
Honor admin settings in user profile Fullname and Email fields

jsperhac:
## Issue summary

Though admin settings are available for controlling end-user access to their profile fields for editing, some of these settings do not work as advertised. (see hub.org/administrator -> Members -> Registration)

Edit access to several end user profile fields can be controlled, at several times. 

Times: 
- account creation
- proxy account creation
- user update on next login
- profile editing

Control settings:
- Required
- Optional
- Hidden
- Read-only

In theory, the admin could disable editing of the user Email by setting it to Required/Required/Readonly/Readonly. However, in the user profile edit interface, a working Edit link was still displayed (and usable). This fix addresses this problem for the Email and Fullname fields.

## This fix

Attempts to set the end-user Email address and Fullname fields as Read Only were unsuccessful. This fix enables:
- administrator can disable editability of Email and Fullname fields in user profile
- Edit link in the UI is hidden if editing is disabled

## Testing

In testing, Admin user can set/unset required/readonly/hidden for these fields and the end user's profile editor enables editing, disables editing, hides field accordingly.

It is likely that some of these other profile fields do not behave ideally. This fix and testing of this fix have been focused on Email and Fullname fields.

Testing was performed on the USP AWS dev instance.

See the very general Jira ticket, USP-57.

This fix by @nkissebe and @jsperhac

### DIFF
--- a/core/plugins/members/profile/views/index/tmpl/default.php
+++ b/core/plugins/members/profile/views/index/tmpl/default.php
@@ -52,6 +52,9 @@ require_once Component::path('com_members') . '/models/incremental/awards.php';
 require_once Component::path('com_members') . '/models/incremental/groups.php';
 require_once Component::path('com_members') . '/models/incremental/options.php';
 
+use Components\Members\Models\Profile\Field;
+
+
 $uid = (int)$this->profile->get('id');
 $incrOpts = new Components\Members\Models\Incremental\Options;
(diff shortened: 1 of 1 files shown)


nkissebe:
I reworked this commit to handle a few other cases and removed some changes that got picked up between 2.2.22 and dev branches. So closing this PR

PR #1632, opened 2023-06-27
Retrieve the resource from the tag plugin to contain the type info. D…

ishunko:
…isplay resource type firs and bolded. Display the result's section (if exists)  on the tags listing page. Fix commet typo.

Partially fixes https://hsrp.sdsu.edu/support/ticket/30 (Displays type of resource for resources and labels groups as such)

### DIFF
--- a/core/plugins/tags/resources/resources.php
+++ b/core/plugins/tags/resources/resources.php
@@ -6,6 +6,8 @@
  */
 
 // No direct access
+use Components\Resources\Models\Entry;
+
 defined('_HZEXEC_') or die();
 
 /**
@@ -434,8 +436,11 @@ class plgTagsResources extends \Hubzero\Plugin\Plugin
 		// Return output
 		return $html;*/
 
-		$row = Components\Resources\Models\Entry::blank()->set($row);
-		$row->set('typetitle', $row->type->get('type'));
+		$query = Entry::all();
+
+		$r = $query->getTableName();
+		$query->whereEquals('id', $row->id);
+		$row = $query->row();
 
 		// Get the component params and merge with resource params
 		$config = Component::params('com_resources');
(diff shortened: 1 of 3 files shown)


core/plugins/tags/resources/resources.php
```diff
@@ -434,8 +436,11 @@ public static function out($row)
 		// Return output
 		return $html;*/
 
-		$row = Components\Resources\Models\Entry::blank()->set($row);
-		$row->set('typetitle', $row->type->get('type'));
+		$query = Entry::all();
+
+		$r = $query->getTableName();
```
nkissebe, line 441: call to getTableName() here is not needed, harmless, but apparently not used

core/plugins/tags/resources/resources.php
```diff
@@ -434,8 +436,11 @@ public static function out($row)
...
+		$query->whereEquals('id', $row->id);
+		$row = $query->row();
```
nkissebe, line 443: Does the original way of fetching the entry row not work right?  Is the effective change here just not setting typetitle to the same value as type?  If that is the effective change why is it needed. I like to add a little doc to changes like this explaining that something was changed/removed. Otherwise someone could come back later seeing this change broke something else and adds it back not knowing the original problem that was trying to be fixed.

core/plugins/tags/resources/resources.php
dbenham: @ishunko - Please remove function if it is not being used

core/plugins/tags/resources/resources.php
```diff
@@ -434,8 +436,11 @@ public static function out($row)
 		// Return output
 		return $html;*/
```
jsperhac, line 437: If this return is unused it should be removed too.

nkissebe (approved):

nkissebe:
Merged and then updated to remove deprecated code in out(). out() used to generate html output directly and modified row to suit. New code is view based and relies on row being output direct from the model.

PR #1645, opened 2023-07-26
Fix de id order

jessewoo:
Can't rely on the order the plugin runs in, so should set the deletion profile key in the controller before calling the plugins, and unset the key after the plugin is ran. Should move the last database updates where we update the field user records until after all plugins run. Should also check plugin returns, shouldn't run the final deletion if any of the plugins fail.

### DIFF
--- a/core/components/com_members/admin/controllers/members.php
+++ b/core/components/com_members/admin/controllers/members.php
@@ -983,10 +983,38 @@ class Members extends AdminController
 	}
 
 	/**
+	 * Different SQL functions
+	 */
+	public function runSelectQuery($query) {
+        $db = \App::get('db');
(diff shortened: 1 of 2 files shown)


dbenham:
Does this need to be in 2.2.26?

nkissebe (approved):

jessewoo:
Hi @dbenham - I think it needs to be with 2.2.26 because it goes with the de-id feature

jessewoo:
@dbenham - wondering if you could merge it because it has to do with de-id feature that is rolled into 2.2.26?

dbenham:
Merged into dev and cherry picked into 2.2.26

PR #1652, opened 2023-08-03
[PURR-125] anonymous ftp workflow - 2 download buttons

JackS9:
- **JIRA task**: https://sdx-sdsc.atlassian.net/browse/PURR-125
- **Support ticket**: https://purr.purdue.edu/support/ticket/2383 
- **PURR core addition**: 8 - anonymous ftp workflow
- **Summary of Issue**: Popular browsers stopped supporting anonymous FTP. Instead, a popup dialog is presented to download an FTP client, which is not convenient. Pascal made backend changes to apache to transfer file via https, and now we need two download buttons to support both types of download.
- **Summary of Fix/Changes**: Created two download buttons: one for HTTPS and another for FTP.
- **Summary of Testing**:  Testing done by PURR staff on their dev environment.  This will be tested again on PURR production hub by PURR staff.
- **Hotfix needed?** This will be pulled to PURR prod as part of testing and targeted for next CMS release (2.2.27)

**NOTE**:  This update contains a migration, but it only provides a default value for param field.

### DIFF
--- a/core/components/com_publications/config/config.xml
+++ b/core/components/com_publications/config/config.xml
@@ -36,6 +36,7 @@
 		<field name="sftptypeblacklist" type="text" menu="hide" default="" label="COM_PUBLICATIONS_CONFIG_SFTP_TYPE_BLACKLIST" description="COM_PUBLICATIONS_CONFIG_SFTP_TYPE_BLACKLIST_DESC" />
 		<field name="documentation" type="text" menu="hide" default="" label="COM_PUBLICATIONS_CONFIG_DOCUMENTATION_PATH" description="COM_PUBLICATIONS_CONFIG_DOCUMENTATION_PATH_DESC" />
 		<field name="deposit_terms" type="text" menu="hide" default="" label="COM_PUBLICATIONS_CONFIG_TERMS_URL" description="COM_PUBLICATIONS_CONFIG_TERMS_URL_DESC" />
+		<field name="ftp_doc" type="text" menu="hide" default="https://purr.purdue.edu/kb/projects/access-datasets-using-ftp-client" label="COM_PUBLICATIONS_CONFIG_FTP_DOC" description="COM_PUBLICATIONS_CONFIG_FTP_DOC_DESC" />
 
 		<field name="include_author_name_in_search" type="list" default="0" label="COM_PUBLICATIONS_CONFIG_SEARCH_AUTHOR_NAME" description="COM_PUBLICATIONS_CONFIG_SEARCH_AUTHOR_NAME_DESC">
 			<option value="0">JNO</option>
--- a/core/components/com_publications/migrations/Migration20220308111111ComPublications.php
+++ b/core/components/com_publications/migrations/Migration20220308111111ComPublications.php
@@ -1,0 +1,54 @@
+<?php
+/**
+ * @package    hubzero-cms
+ * @copyright  Copyright (c) 2005-2022 The Regents of the University of California.
+ * @license    http://opensource.org/licenses/MIT MIT
+ */
+
+use Hubzero\Content\Migration\Base;
+
+// No direct access
+defined('_HZEXEC_') or die();
+
+/**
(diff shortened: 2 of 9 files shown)


core/components/com_publications/config/config.xml
```diff
@@ -36,6 +36,7 @@
 		<field name="sftptypeblacklist" type="text" menu="hide" default="" label="COM_PUBLICATIONS_CONFIG_SFTP_TYPE_BLACKLIST" description="COM_PUBLICATIONS_CONFIG_SFTP_TYPE_BLACKLIST_DESC" />
 		<field name="documentation" type="text" menu="hide" default="" label="COM_PUBLICATIONS_CONFIG_DOCUMENTATION_PATH" description="COM_PUBLICATIONS_CONFIG_DOCUMENTATION_PATH_DESC" />
 		<field name="deposit_terms" type="text" menu="hide" default="" label="COM_PUBLICATIONS_CONFIG_TERMS_URL" description="COM_PUBLICATIONS_CONFIG_TERMS_URL_DESC" />
+		<field name="ftp_doc" type="text" menu="hide" default="https://purr.purdue.edu/kb/projects/access-datasets-using-ftp-client" label="COM_PUBLICATIONS_CONFIG_FTP_DOC" description="COM_PUBLICATIONS_CONFIG_FTP_DOC_DESC" />
```
nkissebe, line 39: Can't have reference to purr.purdue here

core/components/com_publications/migrations/Migration20220308111111ComPublications.php
```diff
@@ -0,0 +1,54 @@
...
+{
+	/**
+	 * Up
+	 **/
+	public function up()
+	{
+		if ($this->db->tableExists('#__extensions'))
+		{			
+			$publicationParams = Component::params('com_publications');
+			$publicationParams->set('ftp_doc', 'https://purr.purdue.edu/kb/projects/access-datasets-using-ftp-client');
```
nkissebe, line 26: can't have reference to purr.purdue here

nkissebe:
Also I don't think we even support ftp at all, regardless of browser support. There might be sftp support but that couldn't really be a button. Maybe a little pop-up to give sftp instructions or something

JackS9:
General comment about these PURR Core Additions: This first pass of PRs is to push the code “as is” from the developer (primarily Jerry Kuang). After that, I can then attempt to fix or make changes based on reviewer comments. However, I think we need to bring the original developer (and others) into the loop if the original issue and/or approach to resolving it are in question.  
Also, my dev environment is not currently set up to test many PURR-related operations - I.e., no real publications or resources, no DOI setup, no curators, no FTP setup, etc.  So some of the suggested changes may need to be developed and tested on dev.purr.purdue.edu first.

dbenham:
I believe the issue is not specific to PURR, but it will need to be reworked in order to be put into core. We *could* consider pointing the button to a page on help.hubzero.org.

core/components/com_publications/config/config.xml
nkissebe: The doc needs to be built into the publications component, then this config will not be needed.

core/components/com_publications/migrations/Migration20220308111111ComPublications.php
nkissebe: this whole thing can be removed if we put doc in publication component

dbenham: @nkissebe - you mean the actual text instead of the link?

dbenham:
I'm going to close the PR out once @kuang5 has had a chance to recreate it in his name. Then we can use the new PR to address  issues going forward. I sent @kuang5 a direct email just now outlining this plan.

dbenham:
Superceeded by PR 1689. Created new PR so Jerry would own the PR instead of Jack (who is no longer with us). I will transfer feedback from this PR to 1689, and once I do, I will close this one out.

PR #1657, opened 2023-08-07
PHP 8.2 backwards compatible fixes

zooley:
This includes a number of backwards compatible fixes for PHP 8.1+, a fix for MySQL 8+/MariaDB10+, and a handful of general bug fixes. This is by no means comprehensive. Some fixes, such as method signatures, were originally PHP8.2+ only but changed to be backwards compatible.


nkissebe (approved):

dbenham:
Seems like a rather fast merge, you sure we can trust the author? ;)

nkissebe:
Shady at best

PR #1658, opened 2023-08-07
Featured groups don't have a div container

jessewoo:
# Reference
- https://sdx-sdsc.atlassian.net/browse/HZ-189
- https://theghub.org/support/ticket/2615

# Summary of the Issue
Need to have feature groups wrapped in a div with class 'groups-container'

# Summary of the Development
Added in outside div container with class named 'groups-container' in 'com_groups > site > views > groups > tmpl > display.php'


jessewoo:
@dbenham @nkissebe

nkissebe (approved):

PR #1662, opened 2023-08-10
[PURR-82] added __isset() to Relational class

JackS9:
### This is an alternate solution to [PR 1661](https://github.com/hubzero/hubzero-cms/pull/1661)

- **JIRA issue**: https://sdx-sdsc.atlassian.net/browse/PURR-82
- **Support ticket**: https://purr.purdue.edu/support/ticket/2439?show=324&search=&limit=20&start=0
- **Summary of issue**: In the publication curation workflow the title of the citation does not appear. In fact the citation is essentially blank.
- **Summary of fix**:  The test for the presence of a citation key (such as title) was incorrectly returning false even though it was set. Added a magic __isset() to core/libraries/Hubzero/Database/Relational.php that checks whether the attribute is set.
- **Summary of testing**:  The issue was replicated on my dev machine and now works with this fix.
- **Hotfix needed?**  Possibly on PURR.

### DIFF
--- a/core/libraries/Hubzero/Database/Relational.php
+++ b/core/libraries/Hubzero/Database/Relational.php
@@ -431,6 +431,17 @@ class Relational implements \IteratorAggregate, \ArrayAccess, \Serializable
 	}
 
 	/**
+	 * Chevk ig attributes (i.e. field) on the model is set
+	 *
+	 * @param   array|string  $key    The key to check if set
+	 * @return  boolean 
+	 */
+	public function __isset($name)
+	{
+		return isset($this->attributes[$name]);
+	}
+
+	/**
 	 * Sets attributes (i.e. fields) on the model
 	 *
 	 * @param   array|string  $key    The key to set, or array of key/value pairs
--- a/core/libraries/Hubzero/Database/Relational.php
+++ b/core/libraries/Hubzero/Database/Relational.php
@@ -431,14 +431,14 @@ class Relational implements \IteratorAggregate, \ArrayAccess, \Serializable
 	}
 
 	/**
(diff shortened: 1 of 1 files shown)


core/libraries/Hubzero/Database/Relational.php
```diff
@@ -430,6 +430,17 @@ public function __get($name)
 		}
 	}
 
+	/**
+	 * Chevk ig attributes (i.e. field) on the model is set
```
nkissebe: Typo in comment Chvk ig

core/libraries/Hubzero/Database/Relational.php
```diff
@@ -430,6 +430,17 @@ public function __get($name)
...
+	 *
+	 * @param   array|string  $key    The key to check if set
```
nkissebe: param can only be string (I know there are similar errors in other functions in this file)

zooley:
Just a note that this functionality already exists on the `Relational` class in the form of the `hasAttribute()` method. Adding an `__isset` magic method isn't a bad idea, though, but should probably just return the other method (or vice-versa).

```
	/**
	 * Check if attributes (i.e. field) on the model is set
	 *
	 * @param  string  $name   The key to check if set
	 * @return  boolean 
	 */
	public function __isset($name)
	{
		return $this->hasAttribute($name);
	}
```

JackS9:
I changed to what Shawn suggested; however, I noticed that __get() could also use hasAttribute()

		// Next check for an attribute on the model
		if (isset($this->attributes[$name]))
		{
			return $this->attributes[$name];
		}

nkissebe (approved):

PR #1663, opened 2023-08-15
[NCN-435] Create plugin plgUserSecret 

jsperhac:
## Summary

This code creates a new plugin suitable for use in `hubzero-cms` core. The plugin creates and maintains a unique 32 character user secret for each Hub user, at login time. If the user has an existing secret, that secret is retained. If the user is ever deidentified, the plugin will null out the user secret. 

User secrets are stored in the `jos_users.secret` column, which is created on up migration of this plugin, and removed on down migration of this plugin. 

## Motivation

This development was undertaken for Nanohub, as part of the Epic ["Salesforce Newsletter Expiration Token Rewrite"](https://sdx-sdsc.atlassian.net/browse/NCN-434), NCN-434. The specific card for this development task is NCN-435.

The goal is to create the secret once per user. The user secret can then be hashed with a unique Hub secret and a unique email campaign secret to create a unique code. This code can be used to form a URL that will be emailed to the user to provide them with access to a Hub newsletter. 

## Code Description

The plugin uses the standard Hubzero plugin architecture, including a migration script that creates/drops the `secret` database column, and a `secret.php` script that provides public `onUserLogin()` and `onUserDeidentify()` functions, as well as several protected functions that generate the secret, check for the secret, and save or null the secret.

## Testing

This plugin was tested on a local Hubzero VM running on CentOS7 under Virtual Box. Tests included: 

- up/down migration to create/drop database column
- login of site and admin users to create non-null secret if needed:
  - on login with null secret column: secret is created
  - on login with non-null secret column: secret is retained
- logout of site and admin users
- simultaneous login of several users
- test of nulling the `secret` database column for a single user

## Deployment

This plugin should be deployed with other changes stemming from Nanohub epic [NCN-434](https://sdx-sdsc.atlassian.net/browse/NCN-434). Hotfixing should not be necessary.

## Revisions

Following code review, the following revisions were completed (as of 14 Sep 23):

- removed composer.json file
- update copyright years to 2023, on all files
- discuss appropriate table location of secret column, with Pascal
- Use Hubzero library function Password::genRandomPassword()  in createUserSecret() 
- improve the README
- indicate where the log() statements in migration script write to (that’s informational muse output)


jsperhac:
One remark: the spec for this epic (specifically card https://sdx-sdsc.atlassian.net/browse/NCN-438) indicates that a given user's secret should be resettable by the admin user. This may require small changes to the original plugin code to enable calling its protected functions (or alternately we might add public functions that enable this operation.) 

I'm not sure which approach is appropriate for a plugin.

core/plugins/user/secret/migrations/Migration20230909000000PlgUserSecret.php
```diff
@@ -0,0 +1,89 @@
...
+	 **/
+	public function up()
+	{
+		$tableName = self::$tableName;
+		$columnData = self::$columnData;
+
+		// Add column to jos_users table to hold user secret
+		// If the table exists, add column if it is absent:
+		if ($this->db->tableExists($tableName))
+		{
```
nkissebe, line 54: Lets talk to Pascal about the secret field being in jos_users table. Sometimes he shares database with people and this might be another field he wants to restrict access to and it might be easier for it not to be in jos_users

core/plugins/user/secret/secret.php
```diff
@@ -0,0 +1,168 @@
...
+
+	/**
+	 * This utility method generates a new user secret.
+	 *
+	 * @return String user secret
+	 */
+	protected function createUserSecret()
+	{
+		// create 32-character secret:
+		$newSecret = shell_exec("head -c 500 /dev/urandom | tr -dc '[:alnum:]' | head -c 32");
```
nkissebe: Lets not use shell_exec() here, we have some functions to generate a random secret/password

core/plugins/user/secret/secret.xml
```diff
@@ -0,0 +1,23 @@
+<?xml version="1.0" encoding="utf-8"?>
+<extension version="2.2" type="plugin" group="user">
+	<name>PLG_USER_SECRET</name>
+	<creationDate>August 2023</creationDate>
+	<author>HUBzero</author>
+	<authorUrl>hubzero.org</authorUrl>
+	<authorEmail>support@hubzero.org</authorEmail>
+	<copyright>Copyright (c) 2005-2020 The Regents of the University of California.</copyright>
```
nkissebe: correct date to 2023

nkissebe: here and elsewhere in this plugin

core/plugins/user/secret/composer.json
```diff
@@ -0,0 +1,24 @@
...
+		"email": "support@hubzero.org",
+		"issues": "https://help.hubzero.org"
+	},
+	"require": {
+		"php" : "^5.4"
+	},
+	"extra": {
+		"install-directory": "core/plugins/user/secret/"
+	}
+}
```
dbenham: Dont think we need this file anymore per our discussion at the group meeting

jsperhac: Got it, this has been removed.

core/plugins/user/secret/secret.xml
jsperhac: This is done.

core/plugins/user/secret/secret.php
jsperhac: Now revised to use Password::genRandomPassword()

core/plugins/user/secret/migrations/Migration20230909000000PlgUserSecret.php
jsperhac: Discussed with Pascal. As a part of the plugin, I have provided nullification of user secret on user identify, which he thinks this is ok.

jsperhac:
@dbenham and @nkissebe, I have made the revisions discussed in code review; all are now present in this PR. Please let me know if you have any questions.

nkissebe:
I hate to bring this up. But since this is going to core and you want to make some UI elements in admin to manipulate it we should probably just implement this directly in core and not as a separate plugin (when would someone ever want to turn it off). Move some of the core functions into the User class (so then easily accessible by admin UI) and add the trigger events to the hubzero user plugin (for lack of knowing how else to add things to event triggers).

dbenham:
Seems a bit late to talk about moving this code from point A to point B, but @nkissebe has a decent argument. We'll let Rich or Amit make the call on if they want to invest the extra time to make this mod, or if we want to go as is. We'll talk about it at the next dev meeting.

dbenham:
Nick and Jeanette will get together to discuss moving this code into a core user.

jsperhac:
Closing this PR. In dev meeting it was decided that this work would be moved into Hubzero Core plugin, plgUserHubzero. A PR for this work will be generated separately. Though the code will be much the same, it will not reside in a freestanding plugin.

PR #1617, opened 2023-04-21
Upon Article Copy, Date Fields Not UTC and Not Updated from Original

jessewoo:
**Source JIRA card(s) and hubzero ticket(s)**
- https://sdx-sdsc.atlassian.net/browse/PLANTSCI-39
- https://plantingscience.org/support/ticket/286

**Brief summary of the issue**
Upon copying an existing article, the create date of the new article is wrong and not the current date. When user changes the start publishing and finish publishing date, the date jumps a few hours back from what was selected on the calendar icon. 

**Brief summary of the fix**
- Upon the copy of the original article, it will be saved with current date for the creation and publishing up date, the alias will also be unique
- The start publishing and finish publishing date will be set at UTC since the calendar selection is in UTC as well. The standardization from all the Hubzero dev will be UTC. 

**Brief summary of your testing**
Manual testing

**Do the change needs to be hotfixed to any production hubs before a normal core rollout**
No

**Double check someone is assigned to review the ticket**
Yes - Nick and David

PR #1664, opened 2023-08-28
[HZ-548] Fix Article "Saving to Copy" Functionality

jessewoo:
- https://sdx-sdsc.atlassian.net/browse/HZ-548 (Bug Fix)
- https://sdx-sdsc.atlassian.net/browse/HZ-534 (Original Content Test Ticket)
- https://github.com/hubzero/hubzero-cms/pull/1617 (Original PR)

Upon copying an existing article, the 'created date' and 'created by' individual doesn't change. The original PR did fix up the start publishing and finish publishing date, which the date remains what was selected. 

**Brief summary of the fix**
- Upon the copy of the original article, the article needs to be SAVED with current date of creation and created by with the current logged in individual. 

- Log in and go to articles library
- Select an article, get to the "edit article" page
- Hit the "Save to Copy" button next to the star button on the right
- The new article should have a new id, created by is set to admin logged in, article alias is null, updated created date, modified date and start publishing date set to now. 
- Update the publishing date and hit the "Save" button, it should be the date selected and doesn't jump around due to time difference in time zone. 

Yes, it needs to be released with 2.2.26 version

Yes - Nick and David

### DIFF
--- a/core/components/com_content/admin/controllers/articles.php
+++ b/core/components/com_content/admin/controllers/articles.php
@@ -440,9 +440,11 @@ class Articles extends AdminController
 		// Upon "saving to copy" or "save to new", blank out alias, update created and published up date to now
         if (in_array($this->_task, $newTasks)) {
             $article->set('created', Date::of('now')->toSql());
+			$article->set('created_by', User::getInstance()->get('id'));
             $article->set('publish_up', Date::of('now')->toSql());
 			$article->set('publish_down', null);
 			$article->set('alias', null);
+			$article->save();
         }
 
 		$this->view
--- a/core/components/com_content/admin/controllers/articles.php
+++ b/core/components/com_content/admin/controllers/articles.php
@@ -438,14 +438,14 @@ class Articles extends AdminController
 		$task = in_array($this->_task, $newTasks) ? 'add' : $this->_task;
 
 		// Upon "saving to copy" or "save to new", blank out alias, update created and published up date to now
-        if (in_array($this->_task, $newTasks)) {
-            $article->set('created', Date::of('now')->toSql());
-            $article->set('created_by', User::getInstance()->get('id'));
-            $article->set('publish_up', Date::of('now')->toSql());
-            $article->set('publish_down', null);
-            $article->set('alias', null);
-            $article->save();
-        }
+		if (in_array($this->_task, $newTasks)) {
+			$article->set('created', Date::of('now')->toSql());
+			$article->set('created_by', User::getInstance()->get('id'));
+			$article->set('publish_up', Date::of('now')->toSql());
+			$article->set('publish_down', null);
+			$article->set('alias', null);
+			$article->save();
+		}
 
 		$this->view
 			->set('task', $task)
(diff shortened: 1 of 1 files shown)


core/components/com_content/admin/controllers/articles.php
```diff
@@ -440,9 +440,11 @@ public function editTask($article = null)
 		// Upon "saving to copy" or "save to new", blank out alias, update created and published up date to now
         if (in_array($this->_task, $newTasks)) {
             $article->set('created', Date::of('now')->toSql());
+			$article->set('created_by', User::getInstance()->get('id'));
             $article->set('publish_up', Date::of('now')->toSql());
 			$article->set('publish_down', null);
 			$article->set('alias', null);
```
nkissebe, line 446: This doesn't seem right. Looks like you are changing these fields on the original article and immediately saving them and then going to the edit view. Seems like these parameters should pass to the edit view and get saved to the new article when the edit view submits.  Perhaps I am missing something.

jessewoo: what was happening was that it took the original author of the author as created_by

dbenham: tabs v spaces issue. Please clean up

jessewoo: reset just my block of code to the indentation of tabs as that was the original file's way of indentation.

dbenham: The way this page was made hurts my brain and made me cry a little inside. 

The "Save to New" functionality isn't a separate function, it piggybacks off the regular save with a special check to add some convoluted logic to a standard save and then redirects to the edit page. So it's messed up already. The "Save to Copy" button creates a new article before bringing up the edit page for the copy, bleh, not the way it should have been done. 

I think Jesse's fix appears odd because the page logic is already odd. Unless we want to invest time in reworking the flow on the page, let's go with this change. Not an ideal fix, but the end result is more accurate to what I would expect a "Save as copy".

dbenham: @nkissebe If you want to push back, I'm ok with that, but I'd reject this fix and create a JIRA card to revisit the flow of the page altogether. This bug isn't worth the time.

I'll let you make the final approval, but if you do reject, you get to reroute the current Jira card (or create another one) to record the need for this fix later.

jessewoo: @dbenham - the bug will persist if not pulled in. So maybe just accept this fix and then make a comment to creating a new JIRA card to visit the page or function. It came up because it's currently affecting a client, and it's production.

jessewoo: This is a bug fix upon a PR that was already accepted in. Upon testing on dev, it didn't work. So this is a fix upon a fix.

dbenham: @jessewoo - What hub is complaining about this?

jessewoo: PlantingScience

jessewoo: There was a thread on Slack -FYI

jessewoo: hot fix in planting science

dbenham (approved):
ok fix for a process that ideally needs refactoring.

PR #1666, opened 2023-08-31
[NCN-502] move youtube link to the online lecture column in course re…

JackS9:
### Adjustment to course lecture presentation for resources type courses

- **JIRA Card**: https://sdx-sdsc.atlassian.net/browse/NCN-502
- **Support Ticket**: https://nanohub.org/support/ticket/378550
This was a wishlist item from NanoHUB for course resource layout. If there is a child resource with a youtube child (logical type youtube), show youtube link in the online lecture column along with Hubpresenter child resources instead of in the supplemental materials column. 
- **Brief summary of the fix/changed code**:  1) Created $logical_alias from alias of child logical_type, similar to how $alias is used for child type, to use strings instead of numeric codes in if and switch statements (much more readable and less prone to changes in numeric codes). 2) Child type aliases of 'elink' and 'youtube' were set to logical_type alias of 'YouTube'.  3) The 'youtube' subtype was then added to the online lecture column instead of defaulting to the supplemental materials column.
- **Brief summary of your testing**: Tested on personal VirtualBox VM with sample resource containing a YouTube link.
- **Needs to be hotfixed to any production hubs?**  No

### DIFF
--- a/core/components/com_resources/site/views/view/tmpl/courses.php
+++ b/core/components/com_resources/site/views/view/tmpl/courses.php
@@ -297,6 +297,7 @@ if ($mode != 'preview')
 								$videoi       = '';
 								$breeze       = '';
 								$hubpresenter = '';
+								$youtube      = '';
 								$pdf          = '';
 								$video        = '';
 								$exercises    = '';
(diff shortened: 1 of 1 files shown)


nkissebe (approved):
I approve. But I am not 100% certain this is the best/proper way to implement but seems good enough.

PR #1669, opened 2023-09-02
[NCN-503] Fixed unit ordering on progress page of course

JackS9:
### Update 20240117-1107 - DRB
Issue NCN-503 cloned from original in Jira, new Jira card: https://sdx-sdsc.atlassian.net/browse/NCN-740
Original ticket is not completely finished

- **JIRA card:** https://sdx-sdsc.atlassian.net/browse/NCN-503
- **Support ticket**: https://nanohub.org/support/ticket/409872
Units in progress bar (graphic) for course were sometimes out of order. Units were being ordered by ID instead of ordering field
- **Brief summary of the fix/changed code**: Added order parameter to units() method in offering model. If true, key in associative array was changed to ordering field instead of id (default key) and used ksort() to reorder array by key (now ordering). Passed order=true to call to offering->units() in student progress view in course progress plugin.
- **Brief summary of your testing**:. Tested and confirmed with course on personal dev machine.
- **Hotfix needed?** no

### DIFF
--- a/core/components/com_courses/models/offering.php
+++ b/core/components/com_courses/models/offering.php
@@ -319,12 +319,10 @@ class Offering extends Base
 	}
 
 	/**
-	 * Get a list of units for an offering
-	 *   Accepts either a numeric array index or a string [id, name]
-	 *   If index, it'll return the entry matching that index in the list
-	 *   If string, it'll return either a list of IDs or names
+	 * Get a list of sections for an offering
 	 *
-	 * @param      mixed $idx Index value
+	 * @param      array   $filters Filters to build query from
+	 * @param      boolean $clear   Force a new dataset?
 	 * @return     array
 	 */
 	public function sections($filters=array(), $clear=false)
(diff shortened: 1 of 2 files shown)


JackS9:
Hold off on approving this.  I'm not convinced these changes address the real problem, since I can't replicate the original issue on my dev machine after returning to a the dev branch.  The changes here appear to do no harm, and they may assure that the units are in proper order, but the units may have already been in the correct order.  The grades themselves may be wrong.

nkissebe (commented):
Deferring to Jack to determine if this fixes an issue, otherwise no objects provided it does so

dbenham:
@JackS9 Confirming why this fix works.

nkissebe:
From Jack:
The “fix” further assures that the units are in proper order, but on closer look they did not appear to actually be out of order.  So the fix does no harm, but it doesn’t address any real issue.  The issue, I believe, is a misunderstanding by the course creator and/or the students in how progress in a course is measured, which is unrelated to scoring of quizzes and exams or to overall grading.

nkissebe:
I am putting this in draft mode for now until we clarify ticket

dbenham:
Pinged the ticket on nanohub to request some clarity.

dbenham:
The JIRA card for this PR needs to be resurrected/recreated, the original fix was not entirely correct. Refer to the nanohub ticket in the description for update on this issue.

nkissebe:
This PR does not appear to address the issue which needs to be made reproducible. There may be an issue with the CMS or the user getting confused by assignments being marked or not being marked as gradeable and counting toward progress. But this will have to be addressed outside this PR and part of a reproduction process. I am going to go ahead and close this PR as we have lost all traction on it.

PR #1671, opened 2023-09-08
[NCN-506] fixed course registration codes generation button not working

JackS9:
### Registration codes for courses not generating
- **JIRA card**: https://sdx-sdsc.atlassian.net/browse/NCN-506
- **Support ticket:** https://nanohub.org/support/ticket/437911
The Generate button on Generate Codes page was not responding. This page is actually a fancy box pop-up window that appears when the Generate button/icon on the Coupon Codes page is clicked, and this page only appears when the codes redeemed number is clicked on within the section page table (Codes column), which only appears when the code generation option is enabled. [Not an obvious path to generate these codes]
- **Brief summary of the fix/changed code**: 
Needed to 1) add Html::behavior('formvalidation) to options view, and 2) close the fancy box after codes generation (and before redirection back to the codes page) in courses.js.
- **Brief summary of your testing**: 
Replicated issue on personal dev machine running latest dev branch and confirmed proper behavior after fix.
- **Hotfixed to any production hubs before a normal core rollout?**  
Possibly on nanoHUB, as there is no current work-around.

### DIFF
--- a/core/components/com_courses/admin/assets/js/courses.js
+++ b/core/components/com_courses/admin/assets/js/courses.js
@@ -58,7 +58,7 @@ jQuery(document).ready(function($){
 
 		$("#btn-generate").on('click', function(e){
 			Hubzero.submitbutton('generate');
-
+			window.parent.$.fancybox.close();
 			window.top.setTimeout("window.parent.location=" + $(this).attr('data-redirect'), 700);
 		});
 
@@ -127,7 +127,7 @@ jQuery(document).ready(function($){
 			changeDynaList(
 				'section_id',
 				offeringsections,
-				document.getElementById('offering_id').options[document.getElementById('offering_id').selectedIndex].value
+				document.getElementById('offering_id').options[document.getElementById('offering_id').selectedIndex].value,
 				0,
 				0
 			);
--- a/core/components/com_courses/admin/views/codes/tmpl/options.php
+++ b/core/components/com_courses/admin/views/codes/tmpl/options.php
@@ -21,6 +21,7 @@ if ($tmpl != 'component')
 }
 
 Html::behavior('framework', true);
+Html::behavior('formvalidation');
 
 $this->js();
 
--- a/core/components/com_publications/admin/controllers/items.php
+++ b/core/components/com_publications/admin/controllers/items.php
@@ -821,6 +821,7 @@ class Items extends AdminController
 			// Update DOI if locally issued
 			if (preg_match("/" . $doiService->_configs->shoulder . "/", $this->model->version->doi))
 			{
+				$doiService->set('authors', $authors);
 				$doiService->update($this->model->version->doi, true);
 
 				if ($doiService->getError())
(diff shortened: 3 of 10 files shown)


dbenham:
FYI: Commit 8a03b08de74a518ecbad6242450d8cd5afaf71b2 was hotfixed to nanohub 2023-09-20-1145

nkissebe:
This PR touches a lot files not relevent to the description. Don't  merge without sorting this out

JackS9:
That one commit that Dave hotfixed is the only thing related to this issue.  I'm not sure how those other unrelated commits got mixed into this branch/PR.

core/components/com_courses/admin/assets/js/courses.js
nkissebe, line 1: Close. The problem is that data-redirect string needs to be enclosed in single quotes. This will cause the page to reload and immediately show the new tokens. WIth just a fancybox close it reverts to the original page which does not have the generated tokens display (until you manuallly reload)

core/components/com_courses/admin/assets/js/courses.js
```diff
@@ -58,7 +58,7 @@ jQuery(document).ready(function($){
 
 		$("#btn-generate").on('click', function(e){
 			Hubzero.submitbutton('generate');
-
+			window.parent.$.fancybox.close();
```
nkissebe, line 61: Close. The problem is that data-redirect string needs to be enclosed in single quotes. This will cause the page to reload and immediately show the new tokens. WIth just a fancybox close it reverts to the original page which does not have the generated tokens display (until you manuallly reload)

core/components/com_courses/admin/assets/js/courses.js
```diff
@@ -127,7 +127,7 @@ jQuery(document).ready(function($){
 			changeDynaList(
 				'section_id',
 				offeringsections,
-				document.getElementById('offering_id').options[document.getElementById('offering_id').selectedIndex].value
+				document.getElementById('offering_id').options[document.getElementById('offering_id').selectedIndex].value,
```
nkissebe, line 130: This correctly fixes the javascript syntax which allows the fix for the generate tokens issue. Unfortuneately this breaks where this function is actually used when adding students to offerings. This is suppossed to reload the sections list when you change the offering but it is returning an empty list. I still have to figure out how this works. Encasing it in if (0) so we can fix the generate token issue without breaking the section listings any more than they already are.

core/components/com_courses/admin/views/codes/tmpl/options.php
```diff
@@ -21,6 +21,7 @@
 }
 
 Html::behavior('framework', true);
+Html::behavior('formvalidation');
```
nkissebe, line 24: This is apparently needed. Not entirely sure of the mechanism but it does in fact fail without it.

core/components/com_courses/admin/views/codes/tmpl/options.php
nkissebe, line 1: Also needs $this->js()->js('timepicker.js'); so don't get js error about missing timepicker function

nkissebe:
Double checked the extra unrelated commits and we already have those accepted via other PRs.

This PR is replaced by manual commit [NCN-619] 1866225b76ad0b3c558c4db5ec3b76045e78749f which applies the relevant commit here along with my suggested fixes.

PR #1672, opened 2023-09-15
[HZ-340] Solr query debug

jsperhac:
## Summary

Provide accessor for Solarium query debug object. The goal is to better enable debugging and development. This change will be largely invisible to users, but useful to developers.

There is no associated Jira card for this change. I've marked it with the Epic for Solr improvements, Phase 1 (HZ-340). 

Hotfixes will not be needed.


core/libraries/Hubzero/Search/QueryInterface.php
```diff
@@ -121,6 +121,14 @@ public function sortBy($field, $direction);
 	 */
 	public function run();
 
+	/**
+	 * getDebug - returns Solarium query debug object
+	 *
+	 * @access public
+	 * @return Solarium\QueryType\Select\Query\Component\Debug
+	 */
+	public function getDebug();
```
nkissebe, line 130: The docblox references a solarium specific return, this interface is supposed to define a generic search interface. Can you think of a way to address this?

dbenham: Jeanette will look into this.

dbenham: @jsperhac - Can you take a look at Nick's last comment and address it? Thanks.

jsperhac: Thanks @dbenham, this fell off my radar. @nkissebe, this is a completely fair question, and a point I had missed.

I suppose the docblock in question (line 128) could be generalized to indicate that some unspecified debug object would be returned. This is a bit unsatisfying to me, though it would achieve generality.

On the other hand, though this is specified as a generic interface, it is used only by Solr, not by the legacy search feature. (That's only a poor argument for failing to keep it general!)

Do you have a suggestion? My concern here is improving logging and debugging, which can be spotty across this codebase. It's possible we would benefit from coming up with some general practices or guidelines for this as we are working on code and the opportunity presents itself. Thank you!

nkissebe: @jsperhac I am content to just change the docblock . Otherwise we need to create some kind of debug class that we implement for each search. We can think about that if its ever needed to be used as such. So just removing the solr string reference is fine with me.

jsperhac: @dbenham and @nkissebe, the docblock has been generalized to indicate a generic debug object rather than a Solarium one.

nkissebe (approved):

core/libraries/Hubzero/Search/QueryInterface.php
dbenham: Looks like the most recent version might not have been properly applied to the PR. @jsperhac will check and resubmit for review.

jsperhac: @dbenham, on second look, this PR is fine. The Solr-specific class ("SolrQueryAdapter") retains a reference to the Solarium object, which is appropriate. The general class ("QueryInterface") correctly shows that it has been generalized to return a debug object. That generalization is what we see in the 4th commit.

All good! This can be merged!

PR #1674, opened 2023-09-15
[HZ-353] Implement Solr for search on Resources landing page

jsperhac:
## Summary

This PR provides the Hubzero Resources landing page with Solr for the search function. It also makes small modifications to the Kimera template so that Solr search results properly display usable filtering for the user. Search results display will include filtering UI elements if the administrator has set up so-called "filter" elements, which is typically used for Resources subtypes (Tools, Teaching Materials, Documents, etc.).

See Resource "filtering" as seen on Nanohub in the screenshot below. The total count of Resources returned is shown at the top, with breakdowns of counts by Resource type. One or more Resource types may be selected using checkboxes, and the filter applied with a button click, which in turn filters the result set displayed.

nh-example

## Motivation

This work is undertaken as part of the Solr enhancement Epic [Phase 1, HZ-340](https://sdx-sdsc.atlassian.net/browse/HZ-40), in particular card [HZ-353](https://sdx-sdsc.atlassian.net/browse/HZ-353).

The objective is to make use of the powerful Solr search engine in preference to our home-grown and outdated Hubzero search functions. Here we roll it out for the Resources landing page.

## Testing

Search on a test Hub was performed from hub.org/search (Solr) and compared to the result set returned from these modifications, in hub.org/resources Search box. These match exactly. The latter was also contrasted with the (homegrown) result set returned from the Resources landing page prior to the deploy of this code branch.

Use of the search in Resources was done with and without the creation of Resources "Filters" of type List (which enumerate the various subtypes of Resources for drilldown in Solr result sets).

When Resources filters were present it was confirmed that Resource subtypes could be selected and filtered on and that the result set displayed changed accordingly.

## Other

No hotfix should be necessary. 

## Out of Scope

Lots more work will be done on Solr as part of this epic.

In particular, with respect to HZ-353, the Resources Browse page must be adapted to accommodate Solr. Wish me luck.


nkissebe (commented):
Jeanette is it feasible to support both solr and internal search depending on site setting? The main search I believe is configurable as such. Just a thought.

jsperhac:
> Jeanette is it feasible to support both solr and internal search depending on site setting? The main search I believe is configurable as such. Just a thought.

Supporting both means of performing searches was not discussed during planning time for this work. It could be done, but would require some backfilling, as presently we have a patchwork: some areas are Solr only (e.g. hubname.org/search) and others are internal search-only. Put another way, there's not really a master switch. 

The job as I understood it was to replace internal search with Solr everywhere practicable. I believe that making both solutions available/selectable would prove more work than it's really worth.

jsperhac:
@nkissebe, the resources landing page now checks the Hub configuration and uses Solr or Basic depending on the setting. This is the same approach taken in `com_search/site/search.php` when determining the controller to use. 

Both the com_resources landing page and the Kimera template override do this check and use the appropriate form, search, and result set display. I've verified that if the config is set to use 'basic' search the old sql-query-based method is used to search the available resources.

jsperhac:
The Resources Browse page is a separate consideration from the Resources landing page and will be a different Jira task and PR (though it could be set up to use the same functionality seen here). 

But, a discussion would help establish what should be delivered. For example: Should adding Solr support to the Hub keep the user interface exactly the same? Should the browse page differ from the landing page? Should search be the same in both cases?

nkissebe:
I manually merged the language fixes with 1884e3ab5f6f2e36f294a9df23ee50634e378a8e

The remainder will have to be resubmitted when we figure out what, if anything we are going to do in general.  I believe the correct solution is maybe to use the hubzero search class which will correctly use which search engine is configured. This should be done the way the main search page does. Maybe. I am honestly not sure. We can resurrect the remainder of this PR as needed.

PR #1685, opened 2023-10-26
Show tagged groups Public Users

jessewoo:
- Previous PR: https://github.com/hubzero/hubzero-cms/pull/1615
- https://sdx-sdsc.atlassian.net/browse/VHUB-9
- https://theghub.org/support/ticket/2601

When a user adds a tag to a group, and upon searching for a tag, that specific group won't appear in the search results for public visitors but will appear for registered and logged in users. 

An example is on the search results of https://theghub.org/tags/ghub. The ISMIP6 group has the associated "Ghub" tag. But that ISMIP6 group only shows up in the results when the user is registered and logged in. The issue is how we can make groups show up for specific tags in the search results for non-registered (public) users.

**Brief summary of the fix**
- The search SQL statement originally filtered out those groups. Talking it over, Nick added conditions to removed the "from" from guests and users in role that view all groups (admin). 

Manual testing locally. It works where group shows up logged in or public. 

No

Yes - Nick and David

### DIFF
--- a/core/plugins/tags/groups/groups.php
+++ b/core/plugins/tags/groups/groups.php
@@ -54,28 +54,40 @@ class plgTagsGroups extends \Hubzero\Plugin\Plugin
 		}
 		$ids = implode(',', $ids);
 
-		$from = '';
+		// The conditions allows tag search results to show groups in public view, not logged in.
+        if (User::isGuest()) {
+            $from = '';
+            $discoverability = " AND a.discoverability=0 ";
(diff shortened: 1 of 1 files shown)


jessewoo:
@dbenham - could we review this in our next dev meeting?

jessewoo:
@nkissebe - thank you for the sql query

jessewoo:
@dbenham - ready to be merged.

dbenham (approved):
Group reviewed on Nov 8th, only feedback was whitespace which was addressed.

PR #1687, opened 2023-11-06
[PURR-89] [PURR-150] rework failure logic in the custom extension repository

dbenham:
- This issue might have multiple related PRs in the system, PURR-89 is only the most recent.
- Custom extension pulls via the admin were not working. The component would blankly report nothing during many scenarios and indicate nothing happening. There were multiple regular expression checks failing due to null values not being checked before doing regex checks.
- I reworked some of the logic that was looking for errors in the php. The code was fragile, doing very specific regex checks for some errors. No custom logic was done with the error detection, other than an attempt to format errors differently than success messages.
- I created a temp custom extension in gitlab, made various updates in that repo and tested this component to make sure it properly grabbed the new code.
- No hotfix changes needed, this code should be part of 2.2.27

Overall, this component needs some major refactoring. It is way over architected. But these fixes should get it up and running for now.

### DIFF
--- a/core/components/com_installer/admin/controllers/customexts.php
+++ b/core/components/com_installer/admin/controllers/customexts.php
@@ -588,7 +588,7 @@ class Customexts extends AdminController
 					$checkoutRepoBranch_response = Cli::call($museCmd, $task='repository');
 					$checkoutRepoBranch_response = json_decode($checkoutRepoBranch_response);
 
-					// did we fail
+					// check for command not found failure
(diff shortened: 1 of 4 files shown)


nkissebe:
Yeah this needs some work, but the changes are in the same style it was originally in and if you say it works better then so be it. I would review the use multistep commands in shell_exec() and the use of echo "" inside of <?php rather than just using <?= to display variables inside of straight html. But both of these cases were in the original code so should not block this PR.

nkissebe (approved):

PR #1689, opened 2023-11-10
[PURR-125] anonymous ftp workflow - 2 download buttons

kuang5:
- JIRA task: https://sdx-sdsc.atlassian.net/browse/PURR-125
- Support ticket: https://purr.purdue.edu/support/ticket/2383
- PURR core addition: 8 - anonymous ftp workflow
- Summary of Issue: Popular browsers stopped supporting anonymous FTP. Instead, a popup dialog is presented to download an FTP client, which is not convenient. Pascal made backend changes to apache to transfer file via https, and now we need two download buttons to support both types of download.
- Summary of Fix/Changes: Created two download buttons: one for HTTPS and another for FTP.
- Summary of Testing: Testing done by PURR staff on their dev environment. This will be tested again on PURR production hub by PURR staff.
- Hotfix needed? This will be pulled to PURR prod as part of testing and targeted for next CMS release (2.2.27)
- NOTE: This update contains a migration, but it only provides a default value for param field.

### DIFF
--- a/core/components/com_publications/config/config.xml
+++ b/core/components/com_publications/config/config.xml
@@ -36,6 +36,7 @@
 		<field name="sftptypeblacklist" type="text" menu="hide" default="" label="COM_PUBLICATIONS_CONFIG_SFTP_TYPE_BLACKLIST" description="COM_PUBLICATIONS_CONFIG_SFTP_TYPE_BLACKLIST_DESC" />
 		<field name="documentation" type="text" menu="hide" default="" label="COM_PUBLICATIONS_CONFIG_DOCUMENTATION_PATH" description="COM_PUBLICATIONS_CONFIG_DOCUMENTATION_PATH_DESC" />
 		<field name="deposit_terms" type="text" menu="hide" default="" label="COM_PUBLICATIONS_CONFIG_TERMS_URL" description="COM_PUBLICATIONS_CONFIG_TERMS_URL_DESC" />
+		<field name="ftp_doc" type="text" menu="hide" default="https://purr.purdue.edu/kb/projects/access-datasets-using-ftp-client" label="COM_PUBLICATIONS_CONFIG_FTP_DOC" description="COM_PUBLICATIONS_CONFIG_FTP_DOC_DESC" />
 
 		<field name="include_author_name_in_search" type="list" default="0" label="COM_PUBLICATIONS_CONFIG_SEARCH_AUTHOR_NAME" description="COM_PUBLICATIONS_CONFIG_SEARCH_AUTHOR_NAME_DESC">
 			<option value="0">JNO</option>
--- a/core/components/com_publications/models/attachments/file.php
+++ b/core/components/com_publications/models/attachments/file.php
@@ -463,7 +463,7 @@ class File extends Base
 	 * @param   boolean  $authorized
 	 * @return  boolean
 	 */
-	public function drawLauncher($element, $elementId, $pub, $blockParams, $elements, $authorized)
+	public function drawLauncher($element, $elementId, $pub, $blockParams, $elements, $authorized, $httpsBtn)
 	{
 		// Get configs
 		$configs = $this->getConfigs($element->params, $elementId, $pub, $blockParams);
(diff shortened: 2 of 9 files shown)


dbenham:
Jerry, 

High level question. Is there any reason using just https downloads for these won't work? What are the typical file sizes download via this link? We can adjust php settings if timeouts are an issue.

dbenham:
In it's current form, this cannot go forward into core. Three options I see:

1) reduce the help document to a javascript popup or tool tip of several sentences, so instructions can easily be included in the interface instead of a separate document 

2) If the help document is a requirement, the document referred to in this fix needs to be created (in it's entirety) via the migration script, so that when this change is sent to other hubs, the document follows it. I'm not sure this is easily done, as the current help document has inline images and a structure that would be a challenge to create via a script, so I'm not even sure if this is a viable option. Using a link so that all hubs point to a centrally located document file is not a viable option, there is no way to ensure the target link exists. 

3) Remove FTP functionality altogether and modify php as needed to support larger downloads via https. i.e. adjust timeouts and memory requirements for the page doing the streaming download.

I'd like to pursue #3, with #1 as a backup option if it turns out file sizes make this option unusable.

core/components/com_publications/models/attachments/file.php
```diff
@@ -572,13 +572,29 @@ public function drawLauncher($element, $elementId, $pub, $blockParams, $elements
...
+								$uri->setPath('app' . DS . $pub->config()->get('sftppath') . DS . $pub->_curationModel->getBundleName(true));
+								$url = $uri->toString();
+							}
+							else
+							{
+								$btnType = "ftp";
+								$label .= ' ' . Lang::txt('with ftp');
+								
+								$uri = \Hubzero\Utility\Uri::getInstance();
+								$uri->setScheme('ftp');
```
dbenham, line 591: This needs to be sftp. I know browsers do not support anonymous ftp, but even if browsers support authenticated ftp, hubzero won't support ftp due to it's fundamentally insecure nature.

kuang5:
Hi Dave,

I accept reducing the help document to a few sentences are good. How about tooltips "Please choose a FTP client such as FileZilla or WinSCP that may be already installed on your computer to download the dataset. For more details look for 'ftp' article in KB"?

We use the anonymous ftp function a lot for both us and users, not only for downloading but also for transferring files to PURR which are going to be imported in project. Please let us know if you plan to disable the function and we need to find out other solution for such kind of use case.

Regarding download large files through https, I'm not sure whether that works for large files, such as the largest dataset on PURR is over 200GB so far, and I think there will be more larger dataset that are going to be published on PURR in the future. 

Jerry

core/components/com_publications/config/config.xml
```diff
@@ -36,6 +36,7 @@
 		<field name="sftptypeblacklist" type="text" menu="hide" default="" label="COM_PUBLICATIONS_CONFIG_SFTP_TYPE_BLACKLIST" description="COM_PUBLICATIONS_CONFIG_SFTP_TYPE_BLACKLIST_DESC" />
 		<field name="documentation" type="text" menu="hide" default="" label="COM_PUBLICATIONS_CONFIG_DOCUMENTATION_PATH" description="COM_PUBLICATIONS_CONFIG_DOCUMENTATION_PATH_DESC" />
 		<field name="deposit_terms" type="text" menu="hide" default="" label="COM_PUBLICATIONS_CONFIG_TERMS_URL" description="COM_PUBLICATIONS_CONFIG_TERMS_URL_DESC" />
+		<field name="ftp_doc" type="text" menu="hide" default="https://purr.purdue.edu/kb/projects/access-datasets-using-ftp-client" label="COM_PUBLICATIONS_CONFIG_FTP_DOC" description="COM_PUBLICATIONS_CONFIG_FTP_DOC_DESC" />
```
dbenham, line 39: Need to make this flexible enough for all hubs. No mention of name/server. Can we make this a small single sentence instructing user to download ftp client. Tell them to google it if they have questions.

dbenham: perhaps a text popup of some sort?

dbenham:
@kuang5 I decided not to include this feature in core, after discussing with the team, we didn't think there was any real limitation to using https. We regularly download large files on other hubs. If we run into any size problems going forward, we're pretty confident we can get around those by tweaking site php settings. 

However, If you want to continue to give PURR this in the interface, you'll have to include it in a custom component. We can talk about the details sometime if you like. The override should be fairly simple.

PR #1683, opened 2023-10-17
[NCN-439] Enable Hub admin to perform per-user secret update

jsperhac:
### Dependencies

This work is part of Epic [NCN-434](https://sdx-sdsc.atlassian.net/browse/NCN-434), whose PRs should all be deployed together:
- #1693, NCN-633 `plgUserHubzero` (on user login, ensure user has a secret in the DB)
- #1683 NCN-439 `com_members` (manage user secret)
- #1676  NCN-437 `com_newsletter` (manage campaign secret)
- #1675 NCN-438 `com_config` (manage hub secret)
- #1695 NCN-702, NCN-440: `com_newsletter` (update access verification)

## Summary

This hubzero-cms core code enables the Hub administrator to reset a Hub member's Secret. The Secret is a 32 character random string unique to each Hub user. 

The changes shown here add a checkbox in the admin interface tab that is used for editing a member's password and related particulars. If the checkbox is checked, and the form is saved, a new user Secret is generated and saved in `jos_users.secret` for that user, replacing any existing Secret.

Navigation: Admin UI -> Users : Members -> select a specific User -> select Password tab -> scroll to bottom of page. 

The following screenshot shows the new checkbox on the Password tab:

admin-members-edit-secret

## Motivation

- Jira Epic: ["Salesforce Newsletter Expiration Token Rewrite"](https://sdx-sdsc.atlassian.net/browse/NCN-434)
- Jira Card: ["Enable Hub admin to perform per-user secret update"](https://sdx-sdsc.atlassian.net/browse/NCN-439)

This development was undertaken for Nanohub, as part of the Epic ["Salesforce Newsletter Expiration Token Rewrite"](https://sdx-sdsc.atlassian.net/browse/NCN-434), NCN-434. The specific card for this work is NCN-439.

Once this epic is deployed, each user on the Hub will have a unique secret. The user secret can then be hashed with a unique Hub secret and a unique Campaign secret to create a unique code. This code can be used to form a URL that will be emailed to the user to provide them with secure, individual access to a newsletter or other materials.

This development will be deployed with other changes from Epic NCN-434 that implement the Campaign secret, the Hub secret, the creation of all User secrets, and their management.

## Code Description

These changes include: 

- a migration script that creates/drops the `jos_users.secret` database column and populates it
- changes to the admin-side form used for password changes, to add the checkbox
- change to the admin members controller to save the checkbox value if activated (==1)
- additions to the language file to provide a label and helptext for the checkbox 

## Testing

[...]

PR #1676, opened 2023-09-22
[NCN-437] Create UI for managing Campaign and its secret

jsperhac:
### Dependencies

This work is part of Epic [NCN-434](https://sdx-sdsc.atlassian.net/browse/NCN-434), whose PRs should all be deployed together:
- #1693, NCN-633 `plgUserHubzero` (on user login, ensure user has a secret in the DB)
- #1683 NCN-439 `com_members` (manage user secret)
- #1676  NCN-437 `com_newsletter` (manage campaign secret)
- #1675 NCN-438 `com_config` (manage hub secret)
- #1695 NCN-702, NCN-440: `com_newsletter` (update access verification)

## Summary

As part of the Nanohub Epic [NCN-434](https://sdx-sdsc.atlassian.net/browse/NCN-434) we will create and maintain individual Campaign secrets (each secret consists of 32 random characters). This PR provides an interface for the Admin user to manage Campaigns, and reset their secrets, if needed.

The Campaign landing page with example records is shown below:

new-campaign-landing-page

## Development

This development was undertaken for Nanohub, as part of the Epic NCN-434, ["Salesforce Newsletter Expiration Token Rewrite"](https://sdx-sdsc.atlassian.net/browse/NCN-434). 

### Epic [NCN-434](https://sdx-sdsc.atlassian.net/browse/NCN-434)

- Original development: "Hub admin campaign management - create admin UI" [NCN-437](https://sdx-sdsc.atlassian.net/browse/NCN-437)
- Revisions: "Updates to PR #1676 following code review" [NCN-618](https://sdx-sdsc.atlassian.net/browse/NCN-618)

## Motivation

The goal of this work is to enable the admin user to manage and reset secrets associated with individual Campaigns. Campaigns might be created for single-click access to secure Hub pages. The campaign information, including the name and description, the secret, the date created, the modification and expiration dates, and the modifying user, is stored in a database table named `campaign`. This Campaign secret can be hashed together with a user's unique secret and a secret unique to the Hub to create a unique code. This code can be used to form a URL that will be emailed to the user (as part of an email campaign) to provide them with access to a Hub newsletter or other resource, without requiring user login. This is general enough functionality to justify its inclusion in core.

Note that resetting the Campaign secret will invalidate any mailings that have utilized it in a hash. 

[...]

PR #1693, opened 2023-12-13
[NCN-633] Rework user secret into core hubzero plugin

jsperhac:
### Dependencies

This work is part of Epic [NCN-434](https://sdx-sdsc.atlassian.net/browse/NCN-434), whose PRs should all be deployed together:
- #1693, NCN-633 `plgUserHubzero` (on user login, ensure user has a secret in the DB)
- #1683 NCN-439 `com_members` (manage user secret)
- #1676  NCN-437 `com_newsletter` (manage campaign secret)
- #1675 NCN-438 `com_config` (manage hub secret)
- #1695 NCN-702, NCN-440: `com_newsletter` (update access verification)

## Summary

This code adds functionality to the existing Hubzero core plugin, `plgUserHubzero`. The new functionality creates and maintains a unique 32 character user secret for each Hub user, at login time. If the user has an existing secret, that secret is retained. If the user is ever deidentified, the plugin will null out the user secret. 

User secrets are stored in the `jos_users.secret` column, which is created on up migration of this plugin. At up migration time, new unique secrets are then generated for all users who have logged in during the past 1 year. The `jos_users.secret` column is removed on down migration of this plugin.

## Motivation

The goal is to create the secret once per user under normal conditions. The user secret can then be hashed with a unique Hub secret and a unique email campaign secret to create a unique code. This code can be used to form a URL that will be emailed to the user to provide them with link-based access to a secure Hub page without requiring login. Should a security incident occur, the user secret can be reset in the admin interface; this is done in `com_members` [PR #1683](https://github.com/hubzero/hubzero-cms/pull/1683/)

## Development

This development was done for Nanohub, as part of the Epic [NCN-434, "Salesforce Newsletter Expiration Token Rewrite"](https://sdx-sdsc.atlassian.net/browse/NCN-434). Details of development task cards:

### Epic: [NCN-434](https://sdx-sdsc.atlassian.net/browse/NCN-434)

- Initial Dev Task: [NCN-435](https://sdx-sdsc.atlassian.net/browse/NCN-435)
- Rework Dev Task: [NCN-633](https://sdx-sdsc.atlassian.net/browse/NCN-633)
- Initial secret population: [NCN-436](https://sdx-sdsc.atlassian.net/browse/NCN-436)
- Migration script secret population: [NCN-693](https://sdx-sdsc.atlassian.net/browse/NCN-693)

## Code Description

The functionality uses the standard Hubzero plugin architecture, including a migration script that creates and populates, or drops, the `secret` database column. The changes here add to the existing plugin's public `onUserLogin()` and `onUserDeidentify()` functions, as well as creating several protected functions that generate the secret, check for the secret, and save or null the secret.

## Testing

This plugin was tested on an AWS Hubzero instance running on CentOS7 and previously on a local VirtualBox Hubzero instance. Tests included: 

- up/down migration to create and populate/drop database column
- login of site and admin users to create non-null secret if needed:
  - on login with null secret column: secret is created
  - on login with non-null secret column: secret is retained
- logout of site and admin users
- simultaneous login of several users
- test of nulling the `secret` database column for a single user

## Deployment

This plugin should be deployed with other changes stemming from Nanohub epic [NCN-434](https://sdx-sdsc.atlassian.net/browse/NCN-434). Hotfixing should not be necessary.

## Revisions

This work was initially developed as a standalone plugin, found in PR [#1663](https://github.com/hubzero/hubzero-cms/pull/1663). 

Following initial code review, these revisions were completed (as of 14 Sep 23):

- removed composer.json file
- update copyright years to 2023, on all files
- discuss appropriate table location of secret column, with Pascal
- Use Hubzero library function Password::genRandomPassword()  in createUserSecret() 
- improve the README
- indicate where the log() statements in migration script write to (that’s informational muse output)

This work was subsequently moved to the `plgUserHubzero` plugin.

### DIFF
--- a/core/plugins/user/hubzero/migrations/Migration20230909000000PlgUserHubzero.php
+++ b/core/plugins/user/hubzero/migrations/Migration20230909000000PlgUserHubzero.php
@@ -1,7 +1,7 @@
 <?php
 /**
  * @package    hubzero-cms
- * @copyright  Copyright (c) 2005-2020 The Regents of the University of California.
+ * @copyright  Copyright (c) 2005-2024 The Regents of the University of California.
  * @license    http://opensource.org/licenses/MIT MIT
  */
 
--- a/core/plugins/user/hubzero/migrations/Migration20230909000000PlgUserHubzero.php
+++ b/core/plugins/user/hubzero/migrations/Migration20230909000000PlgUserHubzero.php
@@ -61,7 +61,7 @@ class Migration20230909000000PlgUserHubzero extends Base
 	/**
 	 * Down
 	 *
-	 * Drop 'secret' column from jos_users table.
+	 * No change is made for down migration.
 	 *
 	 **/
 	public function down()
--- a/core/plugins/user/hubzero/migrations/Migration20230909000000PlgUserHubzero.php
+++ b/core/plugins/user/hubzero/migrations/Migration20230909000000PlgUserHubzero.php
@@ -13,9 +13,9 @@ defined('_HZEXEC_') or die();
 /**
  * Migration script for adding User Secret to plugin
  *
- * This migration script creates a jos_users column for the plugin to populate
- * with a unique user secret, on user login. On 'down' migration the column
- * is dropped from the table.
+ * This migration script checks for jos_users column for the plugin to populate
+ * with a unique user secret, on user login. Creation of this column is done
+ * on up migration in com_members component.
  *
  **/
 class Migration20230909000000PlgUserHubzero extends Base
@@ -24,16 +24,16 @@ class Migration20230909000000PlgUserHubzero extends Base
 	static $tableName = '#__users';
 
 	// specify column to create
-	static $columnData = 	[ ['name' => 'secret',
-								'type' => 'char(32)',
-								'restriction' => 'UNIQUE',
-								'default' => 'NULL'],
-							];
+	static $columnData = [['name' => 'secret',
+		'type' => 'char(32)',
+		'restriction' => 'UNIQUE',
+		'default' => 'NULL'],
+	];
 
 	/**
 	 * Up
 	 *
-	 * Create 'secret' column in jos_users if needed.
+	 * Check for 'secret' column in jos_users table.
 	 *
 	 * Note that we have logging callbacks, per
 	 * https://help.hubzero.org/documentation/22/webdevs/database/migrations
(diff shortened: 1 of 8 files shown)


core/plugins/user/hubzero/migrations/Migration20230909000000PlgUserHubzero.php
```diff
@@ -0,0 +1,130 @@
...
+				$query = $this->_generateSafeAddColumns($tableName, $columnData);
+				$this->_queryIfTableExists($tableName, $query);
+
+			}
+			// Generate and save user secrets
+			$this->generateUserSecrets();
+		}
+
+		// now enable the plugin:
+		$this->enablePlugin('user', 'hubzero');
```
jsperhac: Now that this functionality has been rolled into a fundamental plugin, I think the enablePlugin() and disablePlugin() calls should be removed from this migration script. This migration makes only a small change. plg_user_hubzero will be needed (and should not be disabled) regardless of whether we are using Secrets!

See also the down() function below.

core/plugins/user/hubzero/migrations/Migration20230909000000PlgUserHubzero.php
```diff
@@ -0,0 +1,130 @@
...
+		// generate query to drop column from table:
+		$query = $this->_generateSafeDropColumns($tableName, $columnData);
+
+		// execute the query, if column is found in the specified table:
+		$this->_queryIfTableExists($tableName, $query);
+
+		$this->log('Column `secret` dropped from table `jos_users`', 'info');
+
+		// now disable the plugin:
+		$this->disablePlugin('user', 'hubzero');
```
jsperhac: Now that this functionality has been rolled into a fundamental plugin, I think the enablePlugin() and disablePlugin() calls should be removed from this migration script. This migration makes only a small change. plg_user_hubzero will be needed (and should not be disabled) regardless of whether we are using Secrets!

See also up() function above.

jsperhac:
@dbenham, I removed the enablePlugin() and disablePlugin() calls and rebased this work as discussed. Thank you!

core/plugins/user/hubzero/migrations/Migration20230909000000PlgUserHubzero.php
```diff
@@ -0,0 +1,114 @@
...
+		// Criteria: create secrets for users that have logged in during the last year
+		$criteria = 'DATE_SUB(CURDATE(), INTERVAL 1 YEAR';
+
+		// Fetch ids for all such users who have no secret
+		$targetUsers = \Hubzero\User\User::all()
+						->where('lastvisitdate', '>', $criteria)
+						->whereIsNull('secret');
+
+		// Create and save a unique secret for each such user:
+		foreach ($targetUsers as $oneUser)
```
nkissebe: I think looping through the users is a bad idea, nanohub has hundreds of thousands of users.

It should be possible to do this in a single sql statement and let sql generate a random string. I think I slack'd a few suggestions. I can come up with the statement if you like.

jsperhac: @nkissebe, thank you for your feedback. A couple of points: 
- This loop over a subset of users occurs only on up migration, a one-time operation.
- The up migration loops over all users that have logged in over the last year, as written (this time frame could be changed). For Nanohub, that's about 30k users. One month would be about 3k users.
- Consistent with other parts of this work, a Hubzero library routine is used for secret generation. This follows an earlier recommendation you made.

For new users or very infrequent users, this plugin checks for a not-null user secret at login time, and generates one if needed.

I didn't see a slack from you. If you still have concerns, please post them in this PR so we can keep the discussion is in one place. Thanks!

jsperhac: @nkissebe, I have removed overt table manipulations from the migration script. It is better placed in the com_members migration script. Instead, in the plugin, I check for the existence of the column, and fail with a message ("run the com_members scripts") if column is not found.

Finally, I addressed your concern about updating potentially thousands of user records in the com_members PR, which is found in #1683. The PHP-side loop is replaced by a SQL-side update statement. Thanks for pointing this out.

jsperhac: ### Migration script changes
Removal of column creation from the present PR, and replacement of the loop over users in #1683, is discussed here: https://sdx-sdsc.atlassian.net/browse/NCN-743

dbenham (approved):

jsperhac:
(Superceded by #1704 )

PR #1716, opened 2024-04-30
Repair migration script

jsperhac:
If table exists:
- truncate it
- check its schema, and 
a. If legacy, insert includes `description` and `required` columns
b. If new, insert includes `profile_field_name` column in place of `description` column.


nkissebe (approved):

nkissebe:
We manually applied a corrected version of this

PR #1721, opened 2024-06-10
[CMKC] Email Audits for Review

jessewoo:
# Source JIRA card(s) and hubzero ticket(s)
https://sdx-sdsc.atlassian.net/browse/USP-193

# Brief summary of the issue
PART 2: Audits the different email setTo() of other files
PART 1: https://github.com/hubzero/hubzero-cms/pull/1723

# Brief summary of the fix
Did a grep on other files on core changes, in the email object for those files, updated to be sent Bcc (blind carbon copy) instead of To or Cc. 

# Brief summary of your testing
Have not tried testing on woo.aws.hubzero.org

# Do the change needs to be hotfixed to any production hubs before a normal core rollout
Not at the moment, we can do this when there is more time. 

# Double check someone is assigned to review the ticket
Yes - Nick and David


jessewoo:
Hi @nkissebe - tried removing and adding back setTo, but when I removed the setTo(), without it, I don't receive an email at all. But putting the setTo() back in, I do receive the email.

jessewoo:
able to update the files from grep:

image

jessewoo:
image

PR #1730, opened 2024-08-16
Optimize the code for accessing ORCID and organization id in JSON+LD

kuang5:
PURR Ticket: https://purr.purdue.edu/support/ticket/2557

The ORCID and organization (organization id) for the author in publication are all optional, so the publication submitter might not enter them when submitting a publication. So, it's better to check whether ORCID and organization id (it is acquired by quering organization name on Research Organization Registry) are all available before adding them to JSON+LD metadata set. 

In JSON+LD plugin, add ORCID and organization id to metadata set only when each of them is available.

Accessing a publication page, so the page is accessible. Check head in the source of the page to see if ORCID and organization id is available, they should appear in JSON+LD metadata set.

### DIFF
--- a/core/plugins/publications/jsonld/jsonld.php
+++ b/core/plugins/publications/jsonld/jsonld.php
@@ -104,8 +104,6 @@ class plgPublicationsJsonld extends \Hubzero\Plugin\Plugin
 
 			$givenName = $contributor->givenName;
 			$familyName = $contributor->surname;
-			$orcid = $contributor->orcid;
-			$orgid = $contributor->orgid;
 
 			if (!$givenName)
 			{
@@ -139,9 +137,9 @@ class plgPublicationsJsonld extends \Hubzero\Plugin\Plugin
 				'familyName' => $familyName
 			);
 			
-			if ($orcid)
+			if ($contributor->orcid)
 			{
-				$author['@id'] = "https://orcid.org/" . $orcid;
+				$author['@id'] = "https://orcid.org/" . $contributor->orcid;
 			}
 
 			if ($contributor->organization)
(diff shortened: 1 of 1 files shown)


nkissebe:
Are we sure all the changes made it into this commit. The changes here don't seem to really change anything. Or maybe I am blind, that could be.

kuang5:
Hi Nick,

The changes are minor, but they secure the access to the "orcid" and "orgid", so that PHP won't throw any error when they are or anyone of them is not set as property of the object. It is possible that the publication submitter doesn't enter the orcid and/or organization(orgid is got based on organization name) when submitting publication, which means that "orcid" and/or "orgid" will not be a property of author object. In this case, it will throw error when accessing the publication page, where the JSON+LD is embedded.

Thank you,
Jerry

dbenham:
@nkissebe - was this reply adequate enough to move this forward?

dbenham:
Jerry,

Are you sure this is all the code you need for this? Reading through this one, it doesn't look like your changes affect the functionality in any way, you just cut out two variables and have the code refer directly to properties of the contributor object. Neither Nick or I have a problem with this code, we're just trying to make sure this PR does everything you expect it to. 

If $contributor->orcid or $contributor->orgid is not defined, you'd get an error in the original version of this code and in your new version. Only difference I see is the error line number will just move from line 107->140 and 108 -> 152 depending on what value isn't defined.

kuang5:
Hi Dave,

Yes, the purpose of the code changes are checking the properties orcid and orgid before accessing and using them. The existing code like those two line that are removed will throw error because the properties are not always available for PURR user and publication contributor.

Jerry

nkissebe:
I think I see what is meant. Where those values are used isn't always run but the old code was always trying to reference them at the top of the function.

nkissebe (approved):

PR #1733, opened 2024-09-24
[PURR][#2566]Update the process for setting FOS tag in DOI metadata record

kuang5:
PURR Ticket: https://purr.purdue.edu/support/ticket/2566

Set Field of Science and Technology(FOS) tag in DOI metadata record on DataCite

The process to set the tags of Field of Science and Technology(FOS) in each DOI metadata record on DataCite is changed. The FOS tag is not related to the regular tags anymore, but it is going to be mapped to department tags. Hub administrator can create the FOS tags and department tags in tags component on admin interface, then set the department tag as child of the FOS tag to set up the relationship between department tag and FOS tag.  

Hub administrator is going to set a department tag to the submitted the publication in publication component on admin interface. After saving the changes, the FOS tag is going to be set in the DOI metadata record on DataCite.

Code changes: 
1. Add condition to exclude non-user created tags when getting the tags that user created.
2. Remove the Subject tag setting.
3. Get the FOS tag based on department tag that is set for each publication.
4. In admin controller -> save task, execute the tag saving operation before the operation of updating the DOI metadata record; get the FOS tag before updating the DOI metadata record.

Test
0. Create department tag "Civil Engineering (Dept)" and FOS tag "Civil Engineering (FOS)" in tags component on admin interface. Set department tag "Civil Engineering (Dept)" as child of FOS tag "Civil Engineering (FOS)".
1. Create and submit a publication.
5. Administrator opens the submitted publication in publication component on admin interface, then add department tag "Civil Engineering (Dept)", and finally click the "Save" or "Save&Close" button on upper right.
6. Login on doi.test.datacite.org, find the publication by DOI and check the metadata record to see if FOS tag "Civil Engineering (FOS)" is included. The expected result is that FOS tag "Civil Engineering (FOS)" is set in one of the subject elements in the DOI metadata record.

### DIFF
--- a/core/components/com_publications/admin/controllers/items.php
+++ b/core/components/com_publications/admin/controllers/items.php
@@ -896,14 +896,27 @@ class Items extends AdminController
 				}
 			}
 		}
+		
+		// Incoming tags
+		$tags = Request::getString('tags', '', 'post');
+
+		// Save the tags
+		$rt = new Helpers\Tags($this->database);
+		$rt->tag_object(User::get('id'), $this->model->version->id, $tags, 1, true);
 
 		// Update DOI with latest information
 		if ($this->model->version->doi && !$action)
 		{
-			// Update DOI if locally issued
 			if (preg_match("/" . $doiService->_configs->shoulder . "/", $this->model->version->doi))
 			{
 				$doiService->set('authors', $authors);
+				
+				$fosTag = $this->model->getFOSTag();
+				if (!empty($fosTag))
+				{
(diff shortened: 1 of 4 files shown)


dbenham (approved):
LGTM

Note: 
$tags = Request::getString('tags', '', 'post'); line 901 in core/components/com_publications/admin/controllers/items.php had me concerned initially, but $tags is user input and could be modified, but it isn't used in sql statements, only saved via ORM calls.

PR #1737, opened 2024-10-17
Emulate Flash content using Ruffle

denphi:
This PR aims to restore Flash content on the CMS by integrating the Ruffle emulator. Ruffle is an open-source project that emulates Flash Player, allowing .swf files to be converted into JavaScript on runtime. This conversion enables Flash content to run seamlessly in modern web browsers without the need for the deprecated Flash Player plugin.


core/components/com_resources/site/views/view/tmpl/play.php
```diff
@@ -128,7 +128,14 @@
 			$html .= ' <param name="scale" value="showall" />'."\n";
 			$html .= ' <embed src="'. $url .'" menu="false" quality="best" loop="false" width="100%" height="'.$height.'" scale="showall" name="SlideContent" align="" type="application/x-shockwave-flash" pluginspage="https://www.macromedia.com/go/getflashplayer" swLiveConnect="true"></embed>'."\n";
 			$html .= '</object>'."\n";
-		}
+            $html .= '<script>'."\n";
+            $html .= ' window.RufflePlayer = window.RufflePlayer || {};'."\n";
+            $html .= ' window.RufflePlayer.config = {'."\n";
+            $html .= '   "publicPath": "/core/components/com_resources/site/assets/js/ruffle",'."\n";
```
nkissebe: It would be better if we generate paths without references to core (to allow component overrrides). So this would be generated with somethine like
` Component::path('com_resources') . DS . 'site' . DS . 'assets' . DS . 'js' . DS . 'ruffle' `

core/components/com_resources/site/views/view/tmpl/play.php
```diff
@@ -128,7 +128,14 @@
...
+            $html .= ' }'."\n";
+            $html .= '</script>'."\n";
+            $html .= '<script src="/core/components/com_resources/site/assets/js/ruffle/ruffle.js"></script>'."\n";
```
nkissebe: Same as above

core/components/com_resources/site/views/view/tmpl/play.php
denphi: I Agree, changes committed.

core/components/com_resources/site/views/view/tmpl/play.php
denphi: I Agree, changes committed.

nkissebe (approved):

PR #1746, opened 2024-11-08
Fix getting offset of bool type

toanhminh0412:
This is to deal with the case where `$results` on line 214 is an empty array. Therefore, on line 216, `$results` is falsey, so the entire function will return `false`.

Somewhere in the code, I don't recall where, tries to get an offset from the return value of this function and raises `getting offset of bool type`

### DIFF
--- a/core/plugins/projects/files/files.php
+++ b/core/plugins/projects/files/files.php
@@ -571,7 +571,7 @@ class plgProjectsFiles extends \Hubzero\Plugin\Plugin
 
 			// Set params
 			$used = array();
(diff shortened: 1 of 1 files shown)


nkissebe:
I think for this one the error is in the caller not checking the return value. The return value is intended to be array or bool, so the caller has to check the result.

toanhminh0412:
That's fair. I modify the change to check for the result from the caller now. This is to fix a bug where the dialog for adding files for a new publication doesn't show up properly

PR #1747, opened 2024-11-09
Fix 'replace' being called on some non-compatible data types (int, bool)

toanhminh0412:


nkissebe:
Do you happen to have an example of what triggers the issue you are addressing. I've passed the review of this off to someone else.

toanhminh0412:
Yes I do. This happened when I was adding a new member to a group. On the **Names/Email Addresses** search bar, I typed some characters and there was a dropdown that suggested users that had an username or name matched with the string in the search bar. When I selected a user from the drop down, their name was not displayed properly.

jessewoo:
Hi @toanhminh0412 - could I see a before and after screen capture of the problem and solution? The solution make sense, you used a toString() to display the text of maybe an object.

toanhminh0412:
Oh, I didn't expect this to get merged right away. But for reference, here was the page I got the issue on
Screenshot 2024-11-13 at 9 55 31 AM

It was an AJAX call so the error was in the console:
Screenshot 2024-11-13 at 9 55 38 AM

PR #1748, opened 2024-11-09
Fix a deprecated regex

toanhminh0412:
'-' had a different meaning for PHP 5.6 from PHP 8.2.

This MR makes change to the regex so they behave the same way as before


nkissebe:
I'm not familiar with this change and need to research or otherwise get more info

PR #1752, opened 2024-11-09
Fix php82 deprecated date functions

toanhminh0412:
Replace the deprecated function `strftime`


nkissebe:
I need to verify this returns date in the expected timezone

nkissebe:
Looks like a bug was introduced a while back. Pretty sure this should be

$month = $date->toLocal('m');

If you want to change the PR to this I will accept it or I can make the change outside this PR

toanhminh0412:
I made the change to the PR

PR #1753, opened 2024-11-09
Fix php82 access private protected prop

toanhminh0412:
This MR makes a frequently publicly accessed class attribute public instead of protected

### DIFF
--- a/core/libraries/Hubzero/Component/View.php
+++ b/core/libraries/Hubzero/Component/View.php
@@ -12,6 +12,9 @@ use Hubzero\Document\Assets;
 
 /**
  * Class for a component View
+ * 
+ * @TODO: Change all public attributes of View class to protected
+ * when all references have been fixed to use get()/set()
  */
 class View extends AbstractView
 {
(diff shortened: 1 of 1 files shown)


nkissebe:
I need to review this one. I believe the proper fix will be fix all the instances of this being accessed directly, just need to see how painful that fix would be.

nkissebe:
Yeah this needs to be fixed caller side.  $view->task  should be changed to $view->get('task') to be consistent with intended design (good or bad).

toanhminh0412:
Sure, I want to know what the fix it, too. Because this does seem a bit unlegit

nkissebe:
Ok. Well I don't like it, but for now I think we have to go with this. In fact you need change all the protected attributes to be public in class View. And make a @TODO note that these should someday be changed to protected when all references have been fixed to use get()/set().  I don't understand why this worked in PHP 5.6 (and I think 7.4) or exactly what changed.

toanhminh0412:
I converted all protected attributes to public and added a todo tag

PR #1754, opened 2024-11-09
Fix 'Creation of dynamic property is deprecated'

toanhminh0412:
This MR adds an attribute to a class that had that attribute dynamically created in the code


nkissebe:
I need to review this one more. I have a vague recollection of handling these and I want to be consistent with changes.

PR #1755, opened 2024-11-09
Fix macro not parsed properly issue

toanhminh0412:
This MR fixes a group announcement macro not being parsed properly (It was parsed in the literal string, i.e. `[group.annoucements()]` - something similar to that). After tracing it, turns out it's just how the class was named

### DIFF
--- a/core/plugins/content/formathtml/macros/group/defaulthomepages.php
+++ b/core/plugins/content/formathtml/macros/group/defaulthomepages.php
@@ -14,7 +14,7 @@ use Plugins\Content\Formathtml\Macros\GroupMacro;
 /**
  * Group events Macro
  */
-class DefaultHomePage extends GroupMacro
+class DefaultHomePages extends GroupMacro
 {
 	/**
 	 * Allow macro in partial parsing?


nkissebe:
I'll need to review this one closer, if this is correct the filename should probably be changed as well and corresponding includes references.

nkissebe:
I'll need more information on this. I don't see references to a group.announcement anywhere. And there are no references to a DefaultHomePages only DefaultHomePage so I don't see how this change can help anything and would probably break group home pages in some way.  Can you recreate the error you were fixing or maybe have an old error log.

toanhminh0412:
So here was the exact error I was getting:
Screenshot 2024-11-11 at 10 11 41 PM

After making the fix, here is what I got:
Screenshot 2024-11-11 at 10 12 55 PM

nkissebe:
Hmm that sure seems familiar, I wonder if we fixed that in the main branch. I'll search around.

nkissebe:
I can't recreate. Do you have to populate/configure the group in any particular way to trigger it. I just created a new group and went to the overview view and I don't see an error.

I've seen this error before, but I can't find a reference to it in any messages/mail/notes I have.

nkissebe:
I cherry-picked a fix from the main branch. See if that fixes the issue.
5ee407c4b6db2f32879de112341631d27c84cd0d

toanhminh0412:
The cherry-picked commit works. Nice!

nkissebe:
Closing PR.

PR #1767, opened 2024-11-25
bringing over some PURR changes from branch 2.2.30 back into dev

dbenham:
bringing over some PURR changes from branch 2.2.30 back into dev


nkissebe:
These commits are already in dev

PR #1769, opened 2024-11-26
Removing deprecated code

denphi:
This code is deprecated , http://worldcatlibraries.org is not longer available

### DIFF
--- a/core/components/com_citations/site/controllers/citations.php
+++ b/core/components/com_citations/site/controllers/citations.php
@@ -254,6 +254,7 @@ class Citations extends SiteController
 		// Add some data to our view for form filtering/sorting
 		$this->view->types = Type::all()->rows();
 
+		/*
 		// Get the users id to make lookup
 		$users_ip = Request::ip();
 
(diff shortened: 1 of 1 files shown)


nkissebe:
If you want the code left in and commented out (as you have it now) please add a comment block describing what the commented out code does and why it is commented out.

denphi:
Personally, I would like it to be removed. Every request to https://help.hubzero.org/citations/browse?search=test takes 10 seconds requesting an URL that no longer exists. I believe the original purpose was to load a standard XML file to format the output results, but that file no longer exists. I’m not sure if anyone in the community wants to revive that part of the code, but as I said, I would like it removed.

nkissebe:
Either way is fine with me. Either remove it entirely. Or leave it commented out with a comment explaining why it is commented out so that if anyone goes in looking at the code they know why there is a block commented out.

denphi:
Ok, I just removed commented out code.

nkissebe (approved):

PR #1772, opened 2024-11-27
stop requests when session is done

denphi:
Stop requesting diskusage api calls when sessions are closed /token expired


nkissebe:
remove code rather than comment it out unless you are really unsure about it. in which case document in comment header why you commented out and any reasons you can think it would need to be undone. otherwise just remove the code and ensure the git commit documents the change sufficiently.

nkissebe:
ping

nkissebe (approved):

PR #1782, opened 2025-03-11
Allow template overloading for resources

denphi:
# This change allow template overloading, resources can be assigned to be rendered on a different template defined on its params entry

### DIFF
--- a/core/components/com_resources/config/entry.xml
+++ b/core/components/com_resources/config/entry.xml
@@ -59,5 +59,6 @@
 			<option value="title">Title</option>
 			<option value="author">Author</option>
 		</param>
+		<param name="page_template" type="text" size="20" default="" label="Template" description="name of the overloading template" />
 	</params>
 </extension>
--- a/core/components/com_resources/site/controllers/resources.php
+++ b/core/components/com_resources/site/controllers/resources.php
@@ -1328,6 +1328,12 @@ class Resources extends SiteController
 			App::abort(404, Lang::txt('COM_RESOURCES_RESOURCE_NOT_FOUND'));
 		}
 
+		// Template overloading
+		$page_template = $this->model->params->get("page_template");
+		if ($page_template && $page_template != ""){
+				$app = App::getApplication()->template;
+				$app->template = $page_template;
+		}
 		// Make sure the resource is published and standalone
 		if (!$this->model->get('standalone')) // || !$this->model->isPublished())
 		{


nkissebe:
Tell me what you are trying to achieve exactly, I'm not a big fan of this particular solution and there may be other ways to achieve some of this.

denphi:
Each resource should have an option to assign a specific template.

nkissebe:
Do you have some examples of what you are trying to achieve and why. Again there may be other ways to do this that fit better.

denphi:
as context, every started here:
https://sdx-sdsc.atlassian.net/browse/NCN-334
But Chipshub was created as a different instance.


We asked Amit and then Rich to look for a way to publish content branded as "Chipshub by nanoHUB".
Added to the nanoHUB Project Progress: Development Planning as "Chipshub/nanohub template"
His last comment "Estimate is a bit over a week and speculation is that using the referrer flag will be the best way to do it." 

I agree this is not an ideal solution. Ideally, you should have the ability to create subhubs/subsites in the HZ instance, and link content to the specific subhub/subsite, I think this was at some extent implemented for citations but then only connected to groups.

Or to create a Frankenstein-template that load different folders based on content, but this will be hard to maintain

e.g. https://nanohub.org/tools/st4stars2023/

nkissebe:
And this is different than supergroups?

nkissebe:
Like https://mygeohub.org/

denphi:
similar to a super-group, but we want to have full functionally in the "super-group" (if this make sense) like create/display groups.

e.g. there are resources belonging to the a nanoHUB group, but some of those would be published with the "Chipshub" template

nkissebe:
We can improve on supergroup to an extent, but not that far as far as the bigger picture goes. As far as the resource goes we can probably do something with the layout mechanism which provides ways to use alternative views, but I don't think this was exposed to com_resources. You would put the alternate layout in the site template override directory. It will take a little exploration to see if it will work. Its not an end-user/admin ability, its something a dev/template author would have to add for any alternate resource views you wanted.

nkissebe:
Basically you would have a view file like display_chipshub.php and display_otherhub.php, etc....

denphi:
I tried that, but the layout only affects the component’s rendering, not the overall layout. Feel free to reject this PR. While we search for a better solution, we will have to override.

nkissebe:
You can also change the toplevel "masterfile" that renders everything outside the "component" area (this is the index.php file in the top of the template directory). This is a little easier to do in the PHP 8 branch as I recently was working on some baseline capabilities to support future javascript based components.

PR #1805, opened 2025-06-04
2.4 qubes fixes

mdlama:
@jsperhac Wasn't sure about Newsletter fixes. I _think_ the fix to the controller logic is sound, but I'm not sure about removing the token from the subscribe task. That line wasn't there previously, which allowed the url [https://qubeshub.org/news/newsletter/subscribe](https://qubeshub.org/news/newsletter/subscribe) to work.

### DIFF
--- a/core/components/com_newsletter/models/campaign.php
+++ b/core/components/com_newsletter/models/campaign.php
@@ -127,7 +127,7 @@ class Campaign extends Relational
 		$expiration = $this->get('expire_date');
 
 		$invalidExpiration = empty($expiration);
-		$isExpired = strtotime(Date::of()) > strtotime($expiration);
+		$isExpired = strtotime(Date::of()) > strtotime($expiration ?: '9999-12-31 23:59:59');
 
 		return $invalidExpiration || $isExpired;
 	}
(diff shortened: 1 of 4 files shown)


mdlama:
FYI my comments are in reference to [f98054f](https://github.com/hubzero/hubzero-cms/pull/1805/commits/f98054f14e7cbff6e44239f07fd4dc36cd9ca03c)

nkissebe:
That token would be needed if the subscribe came from a form submission. My guess is maybe wrap a Request::getMethod() == 'POST' check around it. However I am not sure off hand what posts it, I'll have to check the origin of commit that added it

mdlama:
Sounds good. As an additional comment, I don't see anything in the `subscribeTask` function that would require a token, e.g. saving/updating the database. The only request variable it is grabbing is an email string, which it uses to check the database for subscribed newsletters. The rest of the `subscribeTask` function appears to just set up the subscribe view, which is the one that creates the form. Submission of this form seems to call `doSingleSubscribeTask` and `doMultiSubscribeTask` with a token.

nkissebe:
The checktoken is an internal CSRF check. Every form on the site should submit a token and every endpoint that accepts a form should check that a valid token was sent. Prevents form submissions coming from other sites.

nkissebe:
the fact that we also do this with a regular link as well kinda negates this and may be its own issue, but its still something we do

nkissebe (approved):

PR #1814, opened 2025-06-24
Update Ini.php

denphi:
Ignore invalid ini description instead of crashing

### DIFF
--- a/core/libraries/Hubzero/Config/Processor/Ini.php
+++ b/core/libraries/Hubzero/Config/Processor/Ini.php
@@ -77,7 +77,7 @@ class Ini extends Base
 			return false;
 		}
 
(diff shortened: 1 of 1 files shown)


nkissebe:
As a general rule I we aren't going to use the error suppression option anywhere, and even if we did this still gets logged to the error log. If you can give me a URL that triggers the error I can work out how to do the error checks needed.

denphi:
https://nanohub.org/resources/20804/

nkissebe:
Fix for the underlying cause of this issue in bcacba0dd8a8793893644f71a07eb956023a19f4. The data in question was a malformed json string instead of an ini string. I loosened the check for json data to not require a trailing curly brace.

A more proper fix for this issue in general requires setting up an error handler around the call which I do in the main use of parse_ini elsewhere. This is sufficient for now.

PR #1816, opened 2025-06-25
Update list.php

denphi:
I think this is a typo. "%2" -> "%s"


nkissebe:
Yes and no. I need to research how this used to work, I changed how the language parser worked in order to work with a safer ini processing mode. I may have oversimplified things or this may just be a holdover from Joomla, but I see it in a few places.

denphi:
this is related to -> https://nanohub.org/support/ticket/485066

nkissebe (approved):

PR #1821, opened 2025-07-11
Fix file upload for new version of CKEditor

mdlama:
Updated version of CKEditor 4 requires a JSON response after file upload. Additionally, cleaned up the code a little bit.

To trigger the bug: When editing a group page and clicking on the image icon, if the user then clicks on the Upload tab, attaches a file, and then clicks "Send it to the Server", an error is thrown.

Fixes https://qubeshub.org/support/ticket/5085.

### DIFF
--- a/core/components/com_groups/site/controllers/media.php
+++ b/core/components/com_groups/site/controllers/media.php
@@ -431,40 +431,28 @@ class Media extends Base
 		// vars for later
 		$message = '';
 
-		// get request vars
-		$ckeditor     = Request::getString('CKEditor', '', 'get');
-		$ckeditorFunc = Request::getString('CKEditorFuncNum', '', 'get');
+		// do upload
+		$upload = $this->doUpload();
 
-		// if dont have a message we are good to go
-		if ($message == '')
-		{
-			// do upload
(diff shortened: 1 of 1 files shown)


nkissebe:
Thanks for digging into this issue. Any chance you have a ckeditor reference documenting the change, just curious on root cause.

mdlama:
The documentation for the filetools plugin is a broken link at the moment. Here's the File Tools page: https://ckeditor.com/cke4/addon/filetools. If you click on Documentation, you'll see what I mean.

I had to dig directly into the code to see what was going on. Here's the permalink to the bit of code that is affected:

https://github.com/ckeditor/ckeditor4/blob/c7e59ec199298b6b23f4aa7a7668f18572385bac/plugins/filetools/plugin.js#L93

You'll notice that it is expecting a JSON object response, with the old code returning HTML (thereby causing the error).

mdlama:
Wait, here's some documentation: https://ckeditor.com/docs/ckeditor4/latest/guide/dev_file_upload.html#response-file-uploaded-successfully

nkissebe (approved):

PR #1822, opened 2025-07-13
Qubes forum mail fixes

mdlama:
This pull request fixes two bugs.

- ab4570b8a53caf0958a30a0f18bf8aab9bb86e24: Standard null to '' fix
- 5895d1644e9b31edb29dfdef772c6ab51f906941: $value in the setBcc function was somehow null, so set a condition to avoid 

The errors are thrown when attempting to commit a forum post. I _think_ one of the errors (the setBcc one) happens when a hyperlink is included in the post.


nkissebe:
Alternate implementations of the issues addressed here have been committed. Resubmit PR if there is still an issue remaining.

PR #1825, opened 2025-07-28
Update supergroup GitLab to new standards

mdlama:
This PR is needed to update Gitlab supergroup code to new standards implemented post-PHP update. It includes additional refactoring to clean code up a bit, as well as some code to report on Gitlab API errors. Summaries of code changes by file are listed below:

`core/components/com_groups/helpers/gitlab.php`

1. Constructor pulls GitLab config, rather than passing in config parameters
2. Moved free-floating code in manage.php into a validate method to check token
3. Created a general request method to handle general error messages, in particular expired or nonexistent token
4. `_getRequest` method uses `min_access_level=40` instead of `owned=true`. Token must have Maintainer or Owner access role for this to work. Previously, `owned=true` worked as API call was using a GitLab user that owned repositories (for QUBES, user was `qubeshub`). We are now using `oauth2`, however, with token bots that do not own the repositories.

`core/components/com_groups/admin/controllers/manage.php`

1. Use new Gitlab class constructor
2. Use new Gitlab class validate method
3. Add "hub-" prefix to Gitlab Group names
4. Add "sg_" prefix to Gitlab Project names
5. Include more Gitlab API error checking
6. Change remote url to use "oauth2" user
7. Added error check for failed branch protection. Note: The token must have full API access for branch protection to work
8. Added code to update remote url token if remote url token does not match existing token. This is important to avoid having to manually update the remote url when a token expires or changes.

NOTE: For the code changes to work, the Gitlab token needs to have scope "api" and role "Maintainer" or "Owner".

### DIFF
--- a/core/components/com_groups/admin/controllers/manage.php
+++ b/core/components/com_groups/admin/controllers/manage.php
@@ -628,22 +628,8 @@ class Manage extends AdminController
 	 */
 	private function _handSuperGroupGitlab($group)
 	{
-		// get needed config vars
-		$gitlabManagement = $this->config->get('super_gitlab', 0);
-		$gitlabUrl        = $this->config->get('super_gitlab_url', '');
-		$gitlabKey        = $this->config->get('super_gitlab_key', '');
-
-		// do we have repo management on
-		// dont output message
-		if (!$gitlabManagement)
-		{
-			return;
-		}
-
-		// make sure we have a url and key if repot management is on
-		if ($gitlabManagement && ($gitlabUrl == '' || $gitlabKey == ''))
-		{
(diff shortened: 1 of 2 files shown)


mdlama:
FYI: Tested PR on [HubZero DDEV environment](https://github.com/qubeshub/hubzero-docker). Commented out [code to prohibit Gitlab management on non-production server](https://github.com/hubzero/hubzero-cms/blob/2e63f480400de2e0437a09e6ef6ddd7c62068e11/core/components/com_groups/admin/controllers/manage.php#L650) and forced group name to be `hub-qubeshub` to avoid Gitlab Group clutter.

PR #1826, opened 2025-08-05
Dont check groups for anonymous users

denphi:
return empty groups if user is not authenticated

# Before you submit this pull request, please make sure:

- [] Make sure there are links in the PR to the source JIRA card(s) and hubzero ticket(s)
- [] Include a brief summary of the issue **in your own words**
- [] Include a brief summary of the fix/changed code
- [] Include a brief summary of your testing. What did you **specifically** do to investigate that this change has the correct results?
- [] Indicate if the change needs to be hotfixed to any production hubs before a normal core rollout
- [] Double check someone is assigned to review the ticket

**Thanks!**

### DIFF
--- a/core/libraries/Hubzero/User/Helper.php
+++ b/core/libraries/Hubzero/User/Helper.php
@@ -281,10 +281,10 @@ class Helper
 	public static function getGroups($uid, $type='all', $cat = null)
 	{
 		//return empty groups if user is not authenticated
-		if ($uid == 0)
-                {
-                        return false;
-                }
+		if ($uid == 0
+		{
+			return array();
+		}
 		
 		$db = \App::get('db');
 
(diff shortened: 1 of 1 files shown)


nkissebe:
What is the problem being solved?  If anon user I don't think any db queries should be returning anything. At best maybe the early check be returning array() instead of false. But since false is a possible return anyway anything calling it should be checking the value before using it as an array (if that is the problem being solved).

denphi:
During a heavy bot crawling, I saw that anonymous/guest profiles were loading groups, these queries are/should be unnecessary. I modified the return value to match the one returned for empty records (false), but I agree that it should return an empty array instead.

Screenshot 2025-08-05 at 8 39 51 PM

nkissebe (approved):

PR #1828, opened 2025-08-18
Fix tool alias validation

denphi:
PHP8 makes comparison stricter, preg_match returns 0 instead of an empty string.


nkissebe:
Probably should change the condition to !preg_match to catch both no match and error.

denphi:
You are correct; I misread it as returning 0 on error as well. The PR has been modified.

nkissebe (approved):

PR #1837, opened 2025-09-30
Add a switch to control the display of search category

kuang5:
PURR Ticket #2305

User initiated a file publication. When user chooses the "Series/Dataset" in the search category UI in tag section of publication submission process, the publication's type is going to be set to a series publication. 

The change is to add a switch to control the displaying of the search category user interface. By default the switch is set to on. On PURR, we are going to turn it off so the file publication won't be set to a series type since there will be not a radio button for user to choose.

Test procedure:
1. Create a file publication. Navigate to the tag section and you should see the search category user interface that allows you to choose Datasets or Series/Dataset.
2. Login on admin interface and access the configuration of publication component. Set the Search category to off and save the changes.
3. Create a new file publication and navigate to the tag section. You won't see the search category user interface anymore.

### DIFF
--- a/core/plugins/projects/publications/views/draft/tmpl/tags.php
+++ b/core/plugins/projects/publications/views/draft/tmpl/tags.php
@@ -18,6 +18,8 @@ $elName = "tagsPick";
 // Get curator status
 $curatorStatus = $this->pub->_curationModel->getCurationStatus($this->pub, $this->step, 0, 'author');
 
+$searchCategory = $this->pub->config('search_category');
+
 ?>
 
 <!-- Load content selection browser //-->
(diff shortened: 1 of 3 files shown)


core/plugins/projects/publications/views/draft/tmpl/tags.php
nkissebe, line 1: I think you need to add a default value ",1" in the config('search_category') so exist hubs that haven't saved this parameter yet will default to the correct value. Kinda a buggy part of how parameters are created and saved.

kuang5: I have added the default value 1 in the statement below:
$searchCategory = $this->pub->config('search_category', 1);

PR #1844, opened 2025-11-15
avoid loading invalid images

denphi:
avoid loading invalid images

### DIFF
--- a/core/libraries/Hubzero/Image/Processor.php
+++ b/core/libraries/Hubzero/Image/Processor.php
@@ -141,7 +141,7 @@ class Processor extends Obj
 	{
 		try
 		{
-			if (!is_readable($this->source)) {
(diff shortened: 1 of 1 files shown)


nkissebe:
I'll take it as is. But the display function is now more limited in the image file types than it was originally. (see https://www.php.net/manual/en/image.constants.php#constant.imagetype-gif), We'll see if this comes up and add image types if needed.

denphi:
> I'll take it as is. But the display function is now more limited in the image file types than it was originally. (see https://www.php.net/manual/en/image.constants.php#constant.imagetype-gif), We'll see if this comes up and add image types if needed.

I was not sure of committing that restriction until I saw the code was already checking for those: 
https://github.com/hubzero/hubzero-cms/blob/2.4-main/core/libraries/Hubzero/Image/Processor.php#L165

PR #1846, opened 2025-11-18
Update language in plg_projects_publications

kuang5:
We want to change the language "You should expect to hear back from us in the next few days" as attached to "You should expect to hear back from us within 5-10 business days.".
thumbnail_image

Submit a file publication and check whether the language shows as "You should expect to hear back from us within 5-10 business days." on the publication page with dataset's pending approval status.


nkissebe:
Issue was resolved through a language override

PR #1853, opened 2026-01-12
Not all users are added to totals

denphi:
User totals are currently inconsistent in the Members Usage.

At the moment, users who access content only on landing pages (e.g., courses, workshops) appear in the tables but are not included in the aggregated totals, resulting in mismatched numbers within the page.

There should be a single, consistent source of truth. Either: Landing page users are excluded everywhere (tables and totals), or Landing page users are included everywhere.

This PR does the second option.


nkissebe:
Already committed in 657897dfc29037b1b5338eec9903c0f08f1b0160

PR #1854, opened 2026-01-15
Restore truncated url

denphi:
https://nanohub.org/support/ticket/497522

Long URLs are truncated by the VARCHAR(255) `url` column, causing broken redirects even though the full URL exists in the `content` field.

Added a fallback at the redirection point to detect truncated URLs and recover the full URL from the `content` field.


nkissebe:
Already committed in 65c7af7808e54c6a790e300ca700d6b4b2da5829

PR #1870, opened 2026-04-02
Fix com_installer errors reported as success

pascal-meunier:
This is a fix for https://qubeshub.org/support/ticket/5149 where a custom extension was reported as successfully installed but "git pull" had failed internally.
in views/customexts/tmpl/merged.php the list of success and failed messages is handled separately and a success message won't overwrite a failure message.
There are 2 more messages in language/en-GB/en-GB.com_installer.ini depending on how the git pull failed.
In controllers/customexts.php we check for indications of problems: a null value, an empty array (success) and an array of error strings.


pascal-meunier:
urgh, I meant to make a pull request into the 2.4 branch.

PR #1875, opened 2026-04-27
ORCID: Added a guard before rendering the OAuth buttons

pascal-meunier:
ticket https://nanohub.org/support/ticket/505173

Two changes made:                                                                                                                                             
                                                                                                                                                                
  config/config.xml — Changed orcid_service default from "public" to "members". The XML default was "public" but every PHP code path uses 'members' as its      
  fallback, so if an admin ever saved the config page without explicitly choosing a service it would store "public" — and there's no orcid_public_client_id     
  field, making $clientID permanently empty.                                                                                                                    
                                                                                                                                                                
  models/fields/orcid.php — Added an $oauthConfigured guard before rendering the OAuth buttons:                                                                 
  $oauthConfigured = ($srv !== 'public' && !empty($clientID) && !empty($redirectURI));                                                                          
  The "Create or Connect" and "Grant Permission" buttons now only appear when:                                                                                  
  - The service is not 'public' (which is documented as search-only)                                                                                            
  - orcid_members_client_id (or _sandbox_) is actually filled in                                                                                                
  - The redirect URI is configured                                                                                                                              
                                                                                                                                                                
  This prevents users from clicking a broken link that ORCID rejects. Until the admin configures orcid_members_client_id and orcid_members_redirect_uri in      
  Administrator → Components → Members → Options, the buttons will simply not appear.

### DIFF
--- a/core/components/com_members/config/config.xml
+++ b/core/components/com_members/config/config.xml
@@ -270,7 +270,7 @@
 		</field>
 		<field name="rorApiVersion" type="text" menu="hide" default="" label="Research Organization Registry API Version Number" description="The version number of Research Organization Registry API, such as 'v2'." />
 		<field name="@spacer" type="spacer" default="" label="" description="" />
-		<field name="orcid_service" type="list" default="public" label="ORCID Service" description="Select the service to use for ORCID. Sandbox is for testing and debugging purposes. The Public service only allows for searching records, new ID creation is not allowed.">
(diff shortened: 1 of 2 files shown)


pascal-meunier:
config/config.xml is not included in this pull request because:
The following paths are ignored by one of your .gitignore files:
core/components/com_members/config
hint: Use -f if you really want to add them.
hint: Disable this message with "git config advice.addIgnoredFile false"

PR #1876, opened 2026-04-27
Feature/PLG auth us

pascal-meunier:
Motivation: file permissions require a group for people in the US. Filling the iplocation table of the network database with a copy of the ip ranges applicable to the us is silly because that's a lot of rows to copy and it creates a management problem to keep the copy up to date. Instead, this plugin uses an SQL query with a join operation.

Three files created under core/plugins/user/us/:
  - us.php — on login, queries the ipcountry table in the geo database using INET_ATON(). If countrySHORT == 'US', adds the user to location_us. Logs both the
  match and non-match cases. On account deletion, removes the user from location_us (with is_object() guard).
  - us.xml — extension manifest
  - language/en-GB/en-GB.plg_user_us.sys.ini — admin UI label


pascal-meunier:
this PR got different code mixed in. Cancelling and trying again to create a clean PR

PR #1878, opened 2026-04-27
created plugin to add users to a D1 nations group if they login from an IP address from one of those nations

pascal-meunier:
Motivation: file permissions require a group for people from D1 nations. Filling the iplocation table of the network database with a copy of the ip ranges applicable to the D1 nations group is silly because that's a lot of rows to copy and it creates a management problem to keep the copy up to date. Instead, this plugin uses an SQL query with a join operation.

Does plg_user_purdue duplicate what plg_user_geo could do if it was configured to use the purdue group and the "pu" location? From what I can see both use custom locations in the iplocation table of the network database, in this case it would be the "pu" location
[7:09 PM]There's a need to put users from D1 nations into a group at login, so that the tool access limitation done in PHP can be mirrored in the filesystem permissions for /apps by setting UNIX acls using that group

Three files created under core/plugins/user/d1/:
  - d1.php — on login, runs the JOIN query against the geo database. If no row is returned (IP not in any country group), returns immediately without touching group membership. If
  countrygroup == 'D1', adds the user to d1_nation. Any other countrygroup is logged but ignored. On account deletion, removes the user from d1_nation.
  - d1.xml — extension manifest
  - language/en-GB/en-GB.plg_user_d1.sys.ini — admin UI label (edited)

### DIFF
--- a/core/plugins/user/d1/d1.php
+++ b/core/plugins/user/d1/d1.php
@@ -27,11 +27,11 @@ class plgUserD1 extends \Hubzero\Plugin\Plugin
 	 */
 	public function onUserLogin($user, $options = array())
 	{
-		$approved = \Hubzero\User\Group::getInstance('d1_approved');
+		$approved = \Hubzero\User\Group::getInstance('d1_override');
 
 		if (is_object($approved) && $approved->isMember(User::get('id')))
 		{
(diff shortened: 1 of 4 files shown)


pascal-meunier:
New commit: At login, if the user is a member of d1_approved, the plugin now removes them from d1_nation and returns without doing any geo database lookup. The  existing geo-based logic runs only for users not in d1_approved.

PR #1914, opened 2026-08-08
com_saml: replace the hardcoded single-SP IdP with a configurable tru…

denphi:
com_saml could only serve one Service Provider: the trust store was three hardcoded PHP arrays holding a single Canvas entry, and the IdP entity ID, endpoint URLs and certificate paths were nanoHUB-specific constants. The config panel defined params nothing read, and the working site controller existed only as a deployment override.

Replace that with a database-backed trust store and an admin UI, so any number of SPs can be registered without code changes and any hub can use the component as shipped.

  - New tables #__saml_service_providers and #__saml_sessions, with models ServiceProvider and SamlSession.
  - IdP.php becomes a params-driven protocol engine; identity and endpoints derive from the hub's configured URL rather than constants.
  - Core site controller (metadata / login / logout) under /saml/idp/*, so no site override is needed.
  - Admin: SP CRUD, metadata import (paste or fetch), an SSO session audit view with admin-driven cleanup, and a dashboard with certificate health.

Security fixes carried by the same change:

  - Responses go to the SP's registered ACS. The previous code took the ACS from the incoming request, so a forged AuthnRequest naming a trusted issuer could have a signed assertion for the logged-in user delivered to an attacker-controlled URL.
  - Unknown or disabled issuers get a plain 400 instead of a signed Responder message, which was a signing oracle.
  - AuthnRequest signatures are verified against the SP's registered certificate; the metadata already advertised WantAuthnRequestsSigned.
  - SessionIndex is an opaque random value; the hub session ID is no longer disclosed to the SP.
  - Replay protection via UNIQUE (sp_id, request_id), which also closes the check-then-insert race. It is deliberately not cache-based: hubs ship with caching disabled, which would leave the guard silently inert.
  - RSA-SHA256 signing, per-SP NameID and attribute-release policy, optional group restriction, and an audit log.

Behaviour is preserved for the existing Canvas integration: URL shapes are unchanged, the default attribute set matches what was released before, and assertion-level signing stays off by default (the response signature already covers the assertion) behind a per-SP opt-in for SPs that require it.


denphi:
@nkissebe  This is my generalization for Canvas/Instructure.

I’ve extended it to support:

- Entity ID: https://anemoisoftware.com/api/saml2/metadata/
- ACS: https://anemoisoftware.com/api/saml2/acs/ (HTTP-POST)
- SLO: https://anemoisoftware.com/api/saml2/ls/ (HTTP-Redirect)

I’ve tested it on local Hubzero instances, but I haven’t run complete end-to-end tests on the dev or prod environments.

PR #1916, opened 2026-08-10
[PURR][#2793]Fix web page accessibility issues

kuang5:
PURR Ticket: https://purr.purdue.edu/support/ticket/2793

The accessibility issues were reported on Siteimprove. Details are in the ticket problem description and the attached there.

The fixes includes changing the text color to black, setting anchor tag minimum height and width to 44 px, not rendering the ul element that doesn't have any value in item and not public, solved the overuse of italics and all upper cases of text, font size too small, normalize the heading structure, making the input field highlighted when reached by keyboard tab.

The fixes have been tested on production PURR and they all work as expected.

### DIFF
--- a/core/components/com_groups/site/assets/css/groups.css
+++ b/core/components/com_groups/site/assets/css/groups.css
@@ -1111,7 +1111,7 @@
 	font-size: 11px;
 }
 .group-page-toolbar span a {
-	color: #aaa;
+	color: #1c1b20 !important;
 	border: none;
 }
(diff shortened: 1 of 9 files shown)


nkissebe:
Generally I don't like using !important when I can avoid it as it makes managing the cascade difficult. Also we typically don't change the core css, we change the template css. I can work on siteimprove issues.

nkissebe:
I believe these issues have been resolved in other commits now.

